PR #6, opened 2016-06-22
Replace Array.prototype.find for IE9+ Browser support

### LINKED ISSUE #4: IE 9+ Support
Hi David,
just a brief remark: According to your README file you are supporting IE9+. I noticed that since version 1.0.3 you are now using the Array.prototype.find Method in [1]. This will break in IE <= 11.

[1] [index.js#L53](https://github.com/davidtheclark/tabbable/blob/master/index.js#L53)

knoefel:
Fixes #4

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,9 +52,12 @@ function createIsHidden() {
 
     var cached;
 
-    nodeCache.some(function(item) {
-      return item[0] === node && (cached = item);
-    });
+    for (var i = 0, length = nodeCache.length; i < length; i++) {
+      if (nodeCache[length] === node) {
+        cached = node;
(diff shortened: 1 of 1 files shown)


davidtheclark:
Thanks for the PR! I'd rather use a `for`-loop that just stops after the first match, instead of a heretical use of `some()` :)

knoefel:
I replaced the `some()` method as requested. Altough i found the `some()` implementation much more readable. You had chosen an Array method in the first place, so i thought i stick to it. But i guess it's just a matter of taste.

davidtheclark:
Published in 1.0.4

PR #11, opened 2017-03-31
Add alternative for obtaining tabIndex for SVGs

iandevlin:
MS Edge and IE 11 have issues with SVGs whereby they make them tabbable even if their containing element has a `tabindex` of `-1 `or isn't even tabbable at all.

Adding `focusable="false"` and `tabindex="-1"` usually solves this, but `tabbable` is incorrectly adding them to its list of `tabbables`.

Calling `candidate.tabIndex` on an SVG returns `undefined`, but using `getAttribute('tabindex')`works fine.

This PR provides this as a fallback.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ module.exports = function(el) {
   var candidate, candidateIndex;
   for (var i = 0, l = candidates.length; i < l; i++) {
     candidate = candidates[i];
-    candidateIndex = candidate.tabIndex;
+    candidateIndex = candidate.tabIndex || candidate.getAttribute('tabindex');
 
     if (
       candidateIndex < 0
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ module.exports = function(el) {
   var candidate, candidateIndex;
   for (var i = 0, l = candidates.length; i < l; i++) {
     candidate = candidates[i];
-    candidateIndex = candidate.tabIndex || candidate.getAttribute('tabindex');
+    candidateIndex = parseInt(candidate.getAttribute('tabindex'), 10) || candidate.tabIndex;
 
     if (
       candidateIndex < 0


iandevlin:
Hmm ok, all the tests fail, I am not sure why though. Will try to find out.

index.js
```diff
@@ -20,7 +20,7 @@ module.exports = function(el) {
   var candidate, candidateIndex;
   for (var i = 0, l = candidates.length; i < l; i++) {
     candidate = candidates[i];
-    candidateIndex = candidate.tabIndex;
+    candidateIndex = parseInt(candidate.getAttribute('tabindex'), 10) || candidate.tabIndex;
```
davidtheclark: What if you set `focusable="false"` but not `tabindex`? I've read that `focusable="false"` is the only way to really prevent this behavior: https://stackoverflow.com/questions/18646111/disable-onfocus-event-for-svg-element.

iandevlin: I have found that I need both to prevent this behaviour. Regardless, SVGs with a `tabindex` of `-1` shouldn't be identified as a `tabbable`.

iandevlin:
Hey! Is this going to be merged any time soon, or a different fix implemented? I kinda need it, hence I am being annoying. Sorry.

index.js
davidtheclark: I have found that I only need `focusable="false"` --- curious why you're seeing something different. Nothing in this PR addresses *that*, tright?

iandevlin: No, nor should it? Or do you think that it should?

davidtheclark:
Published as 1.0.6. Sorry for the delay, and thanks for the contribution, @iandevlin!

iandevlin:
Excellent! Thank you!

PR #12, opened 2017-08-25
Use original dom position index to stabilize sort func

aceysmith:
On some browsers (Chrome specifically) Array.prototype.sort is not stable. This causes the order of like-tabindexed elements to be unexpected in some cases.

By comparing the array index of elements from the querySelectorAll, we can tie-break elements with the same tabindex.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@ module.exports = function(el) {
       basicTabbables.push(candidate);
     } else {
       orderedTabbables.push({
+        index: i,
         tabIndex: candidateIndex,
         node: candidate,
       });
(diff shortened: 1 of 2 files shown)


davidtheclark:
@aceysmith: The change looks good. Thanks for looking into it! Do you think there's any way we could test this?

aceysmith:
I certainly thought about that. IIRC your test driver uses Firefox which uses a stable sort for Array, so the test wouldn’t be able to verify it did the right thing, only that it didn’t do the wrong thing. But yeah I can certainly put something together to at least make sure it doesn’t affect the order for Firefox adversely.

davidtheclark:
@aceysmith: If a path seems clear to you, it's probably worth it. If not, the code looks fine and the existing tests pass ... so I'm not too worried. Your call 👍

aceysmith:
Alrighty! I enhanced the non-linear test case instead of creating another as it seems to have a similar intent. Let me know if you have any other concerns. If not, it's good to go on my end!

davidtheclark:
Thanks again, @aceysmith. Published as 1.0.7.

PR #13, opened 2017-09-01
Feature/add iframe support

white-water:
This enables `tabbable` to be invoked on elements of an iframe (as well as elements of the main window) and use the correct document context based on the subject element.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,11 @@
 module.exports = function(el) {
+  var elementDocument = el.ownerDocument;
   var basicTabbables = [];
   var orderedTabbables = [];
 
   // A node is "available" if
   // - it's computed style
-  var isUnavailable = createIsUnavailable();
+  var isUnavailable = createIsUnavailable(elementDocument);
 
   var candidateSelectors = [
     'input',
@@ -26,7 +27,7 @@ module.exports = function(el) {
       candidateIndex < 0
       || (candidate.tagName === 'INPUT' && candidate.type === 'hidden')
       || candidate.disabled
-      || isUnavailable(candidate)
+      || isUnavailable(candidate, elementDocument)
     ) {
       continue;
     }
@@ -54,7 +55,7 @@ module.exports = function(el) {
   return tabbableNodes;
 }
 
-function createIsUnavailable() {
+function createIsUnavailable(elementDocument) {
   // Node cache must be refreshed on every check, in case
   // the content of the element has changed
   var isOffCache = [];
@@ -65,14 +66,14 @@ function createIsUnavailable() {
   // "off" state, so we need to recursively check parents.
 
   function isOff(node, nodeComputedStyle) {
-    if (node === document.documentElement) return false;
+    if (node === elementDocument.documentElement) return false;
 
     // Find the cached node (Array.prototype.find not available in IE9)
     for (var i = 0, length = isOffCache.length; i < length; i++) {
       if (isOffCache[i][0] === node) return isOffCache[i][1];
     }
 
-    nodeComputedStyle = nodeComputedStyle || window.getComputedStyle(node);
+    nodeComputedStyle = nodeComputedStyle || elementDocument.defaultView.getComputedStyle(node);
 
     var result = false;
 
@@ -87,10 +88,10 @@ function createIsUnavailable() {
     return result;
   }
 
-  return function isUnavailable(node) {
-    if (node === document.documentElement) return false;
+  return function isUnavailable(node, elementDocument) {
+    if (node === elementDocument.documentElement) return false;
 
-    var computedStyle = window.getComputedStyle(node);
+    var computedStyle = elementDocument.defaultView.getComputedStyle(node);
 
     if (isOff(node, computedStyle)) return true;
 
--- a/index.js
+++ b/index.js
@@ -36,6 +36,7 @@ module.exports = function(el) {
       basicTabbables.push(candidate);
     } else {
       orderedTabbables.push({
+        index: i,
         tabIndex: candidateIndex,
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -88,10 +89,10 @@ function createIsUnavailable() {
     return result;
   }
 
-  return function isUnavailable(node) {
-    if (node === document.documentElement) return false;
+  return function isUnavailable(node, elementDocument) {
```
davidtheclark: If `elementDocument` is already available in the closure, because it's passed to `createIsUnavailable`, it's not necessary here, it is?

test/tabbable.test.js
```diff
@@ -13,18 +13,36 @@ var fixtures = {
...
   var html = fixtures[fixtureName];
-  var root = document.createElement('div');
+  var root = doc.createElement('div');
   root.innerHTML = html;
-  document.body.appendChild(root);
+  doc.body.appendChild(root);
+  return root;
+}
+
+function _getTabbableIds(node) {
```
davidtheclark: I don't think there's any need for this leading underscore convention, because none of these functions are part of a public API — right?

test/tabbable.test.js
```diff
@@ -40,117 +58,125 @@ describe('tabbable', function() {
...
-    var actual = fixture('nested').getTabbableIds();
-    var expected = [
-      'tabindex-div-2',
-      'tabindex-div-0',
-      'input',
-    ];
-    assert.deepEqual(actual, expected);
-  });
+    describe(assertionSet.name, function() {
+    it('basic', function() {
```
davidtheclark: Indentation seems off here.

test/tabbable.test.js
```diff
@@ -40,117 +58,125 @@ describe('tabbable', function() {
...
-      'href-anchor',
-      'textarea',
-      'button',
-      'tabindex-div',
-      'hiddenParentVisible-button',
-    ];
-    assert.deepEqual(actual, expected);
-  });
+  [
+    { name: 'window', func: fixture },
```
davidtheclark: How about `getFixture` instead of `func`, since `func` is not descriptive?

test/tabbable.test.js
```diff
@@ -40,117 +58,125 @@ describe('tabbable', function() {
...
-    var expected = [
-      'tabindex-div-2',
-      'tabindex-div-0',
-      'input',
-    ];
-    assert.deepEqual(actual, expected);
-  });
+    describe(assertionSet.name, function() {
+    it('basic', function() {
+        var actual = fixture('basic').getTabbableIds();
```
davidtheclark: Shouldn't this be using the function `assertionSet.getFixture` (or `func`)?

test/tabbable.test.js
```diff
@@ -40,117 +58,125 @@ describe('tabbable', function() {
...
+          'visible-button-2',
+          'visible-button-3',
+          'initially-hidden-button-1',
+          'initially-hidden-button-2',
+        ];
+        assert.deepEqual(actualB, expectedB);
+      });
+    });
   });
 });
```
davidtheclark: As part of the PR, can we add an entry to the changelog and some content to the documentation?

index.js
white-water: Excellent point, thank you!

test/tabbable.test.js
white-water: No problem, I'm happy to remove the underscores.

test/tabbable.test.js
white-water: Good call, I'll go with `getFixture` then.  And, after looking at this comment, I came to the same realization you pointed out below: I'd failed to actually use this parameterized function!

test/tabbable.test.js
white-water: It sure is, thank you!

test/tabbable.test.js
white-water: Yup, thanks for catching this too.  (This is what I get for trying to rush through this PR on Friday!)  I'll have an update up shortly!

test/tabbable.test.js
white-water: Sure thing, I'll add something to `CHANGELOG.md` for version `1.0.8` and describe the iframe support addition—did you have any additional documentation targets in mind?

aceysmith:
I’m sure white-water will chime in here too, but we’ve been using this in a production environment for some time. We just recently decided to submit these changes (and the sort stability I PR’d for last week) upstream.

white-water:
@davidtheclark excellent question, and, seconding what @aceysmith noted, we've actually been using this patch in production for several months.

davidtheclark:
Thanks again @white-water @aceysmith. This is a good lesson for me in writing code that can better handle iframes,

white-water:
Thanks, @davidtheclark!

davidtheclark:
@white-water @aceysmith I have a follow-up question for you: After the changes you made here (`document` -> `element.ownerDocument`, `window` -> `document.defaultView`) the only ones that are really important for writing browser JS that works within iframes? Or are there are other important features you know about that aren't applicable to this library? Do you have any suggested reading material about this?

I ask because I want to learn how to write more iframe-friendly code by default, and I'm having trouble finding a list of things to look out for.

aceysmith:
I suppose https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe + https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement might be good references, it outlines differences in contentWindow/contentDocument vs window/document in the Scripting section, for instance. But as far as a guide giving more of an exposition, I don't know of anything better than what google probably does.

Issue #14, opened 2017-09-10
Option to check a container too


vonagam:
Right now only nodes within a container get tested, but not a container itself. I have use case when i want a container to be possibly included in resulted array.

In current state i may get what i need, by retrieving tabbables of a parent of a container and then filtering results by inclusion of container. My concern is that of performance. Siblings of a container can be big, while container itself may be small or empty, it may result in much of unnecessary work.

Another way is to add option which will tell to prepend a container to candidates list if it matches one of selectors.

What are your opinion on adding such option? It is ok or is it out of scope of this library?

vonagam:
Just a note for other people, who may need to customize their querying (like in #10):

Found a third, hacky way. Basically right now tabbable takes one argument `el`. And from this supposed element it needs only two things: `ownerDocument ` and `querySelectorAll`. So, you can just pass not an element, but an object with those keys, where `querySelectorAll` implements custom querying.

davidtheclark:
@vonagam: I understand the use case. How about adding an option `includeContainer: boolean`? There aren't any options yet, but we could add a second argument that's an options object. I can't think of any good reason *not* to. I think that would solve your problem, right?

vonagam:
Yes. That would be perfect.

davidtheclark:
👍  open to a PR to introduce this option.

davidtheclark:
Closed by #15.

PR #15, opened 2017-09-17
Add includeContainer option

vonagam:
Here is PR for #14.

Uses [matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) and [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) both support IE9.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,6 @@
-module.exports = function(el) {
+module.exports = function(el, options) {
+  options = options || {};
+
   var elementDocument = el.ownerDocument;
   var basicTabbables = [];
   var orderedTabbables = [];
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -18,6 +20,18 @@ module.exports = function(el) {
...
 
+  if (options.includeContainer) {
+    var matches = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
+
+    if (
+      candidateSelectors.some(function(candidateSelector) {
+        return matches.call(el, candidateSelector);
+      })
+    ) {
+      candidates.unshift(el);
```
davidtheclark: Wouldn't we want to push this directly into `orderedTabbables`?

index.js
```diff
@@ -18,6 +20,18 @@ module.exports = function(el) {
 
   var candidates = el.querySelectorAll(candidateSelectors);
 
+  if (options.includeContainer) {
+    var matches = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
```
davidtheclark: Interesting! I didn't know about this function.

index.js
vonagam: Well, for my use case - no. I wanted to apply same rules to container. So if it has tabindex -1 or disabled or hidden it should not get in resulted list.

(I see the problem though that querySelectorAll returns not array, but a NodeList which does not have methods to add item to it)

vonagam: And if it has tabindex = 0, then if we push it into `orderedTabbables` then it will appear after all `basicTabbables` but before any other `orderedTabbables`, which is incorrect result.

vonagam:
Here, reused "nested" fixture for the test. Added note about options in Usage section.

davidtheclark:
Thanks again, @vonagam. Published in 1.1.0.

PR #17, opened 2017-11-02
Fix crash when tabbable called with document

felipeochoa:
#13 broke `tabbable(document)` since `ownerDocument` is `null` on all document nodes.[1] This PR fixes that calling pattern

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 module.exports = function(el, options) {
   options = options || {};
 
-  var elementDocument = el.ownerDocument;
+  var elementDocument = el.ownerDocument || el;
   var basicTabbables = [];
   var orderedTabbables = [];
 
--- a/test/tabbable.test.js
+++ b/test/tabbable.test.js
@@ -46,6 +46,15 @@ function fixtureWithIframe(fixtureName) {
   };
 }
 
+function fixtureWithDocument(fixtureName) {
+    var root = createRootNode(document, fixtureName);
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -1,7 +1,7 @@
 module.exports = function(el, options) {
   options = options || {};
 
-  var elementDocument = el.ownerDocument;
+  var elementDocument = el.ownerDocument || el;
```
davidtheclark: Are we certain that `document` is the only Node that does not have `ownerDocument`? Or should we check `el instanceof Document ? el : el.ownerDocument`?

felipeochoa: From [W3C DOM4](https://www.w3.org/TR/dom/#dom-node-ownerdocument):

> Each **node** has an associated **node document**, set upon creation, that is a **document**. 
> [...]
> The **ownerDocument** attribute must run these steps:
> 1. If the **context object** is a **document**, return null.
> 2. Return the **node document**. 

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) indicates that all browsers are compatible with this API

davidtheclark: Awesome, thanks @felipeochoa.

davidtheclark:
@felipeochoa Published in 1.1.1

felipeochoa:
Awesome, thanks!

PR #20, opened 2017-11-29
Support Web Components

DRiFTy17:
A basic description of these changes is that I replaced the standard querySelectorAll function with my own implementation of it that also traverses shadow roots/slots, but otherwise it _should_ function just as it did before if no shadow roots are found. Let me know if you see any issues with supporting various browsers that I may have missed, but things appear to checkout on all common browsers from what I can tell.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,8 +2,7 @@
   "env": {
     "node": true,
     "browser": true,
-    "mocha": true,
-    "es6": true
+    "mocha": true
   },
   "extends": ["eslint:recommended"],
   "rules": {
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@
   "env": {
     "node": true,
     "browser": true,
-    "mocha": true
+    "mocha": true,
+    "es6": true
   },
   "extends": ["eslint:recommended"],
   "rules": {
--- a/index.js
+++ b/index.js
@@ -95,11 +95,15 @@ function deepQuerySelectorAll(rootElement, selectors, checkRootElement) {
 }
 
 function matchesSelectors(el, selectors) {
-  if (el.nodeType === Node.TEXT_NODE) {
+  try {
+    if (el.nodeType === Node.TEXT_NODE) {
+      return false;
+    }
+    var matchesFn = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
+    return selectors.some(function(selector) { return matchesFn.call(el, selector); });
+  } catch (e) {
     return false;
   }
-  var matchesFn = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
-  return selectors.some(function(selector) { return matchesFn.call(el, selector); });
 }
 
 function createIsUnavailable(elementDocument) {
--- a/index.js
+++ b/index.js
@@ -1,3 +1,5 @@
+var shadowUtils = require('shadow-utils').ShadowUtils;
+
 module.exports = function(el, options) {
   options = options || {};
 
@@ -20,7 +22,7 @@ module.exports = function(el, options) {
 
   var candidates;
   if (options.deep) {
-    candidates = deepQuerySelectorAll(el, candidateSelectors, options.includeContainer);
+    candidates = shadowUtils.deepQuerySelectorAll(el, candidateSelectors, options.includeContainer);
   } else {
     candidates = el.querySelectorAll(candidateSelectors.join(','));
 
@@ -74,51 +76,6 @@ module.exports = function(el, options) {
   Array.prototype.push.apply(tabbableNodes, basicTabbables);
 
   return tabbableNodes;
-}
-
-/**
- * Walks the DOM tree starting at a root element and checks if any of its children 
- * match the provided selectors. Similar to the native `querySelectorAll` except
- * that it will traverse the shadow DOM as well as slotted nodes.
- * @param {Element} rootElement The element to start querying from.
- * @param {string[]} selectors An array of CSS selectors.
- * @param {boolean} checkRootElement True if the provided root element is to be matched against the selectors.
- */
-function deepQuerySelectorAll(rootElement, selectors, checkRootElement) {
-  var nodes = [];
-
-  if (checkRootElement) {
-    if (matchesSelectors(rootElement, selectors) && nodes.indexOf(rootElement) === -1) {
-      nodes.push(rootElement);
-    }
-  }
-
-  if (rootElement.tagName === 'SLOT') {
-    var slotNodes = rootElement.assignedNodes();
-    slotNodes.forEach(function(slottedNode) {
-      nodes = nodes.concat(deepQuerySelectorAll(slottedNode, selectors, true));
-    });
-  } else {
-    if (rootElement.shadowRoot) {
-      rootElement = rootElement.shadowRoot;
-    }
-
-    var node = rootElement.firstElementChild;
-    while (node) {
-      nodes = nodes.concat(deepQuerySelectorAll(node, selectors, true));
-      node = node.nextElementSibling;
-    }
-  }
-
-  return nodes;
-}
-
-function matchesSelectors(el, selectors) {
-  if (el.nodeType !== Node.ELEMENT_NODE) {
-    return false;
-  }
-  var matchesFn = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
-  return selectors.some(function(selector) { return matchesFn.call(el, selector); });
 }
 
 function createIsUnavailable(elementDocument) {
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ module.exports = function(el, options) {
     }
 
     if (candidate.tagName === 'SLOT') {
-      var slotCandidates = candidate.assignedNodes();
+      var slotCandidates = candidate.assignedNodes().filter(function(node) { return node.nodeType !== Node.TEXT_NODE; });
       var slotChildCandidates = [];
 
       slotCandidates.forEach(function(node) {
@@ -62,7 +62,9 @@ module.exports = function(el, options) {
         }
       });
 
-      slotCandidates = slotCandidates.concat(slotChildCandidates);
+      if (slotChildCandidates.length) {
+        slotCandidates = slotCandidates.concat(slotChildCandidates);
+      }
 
       slotCandidates = slotCandidates.filter(function(node) {
         return candidateSelectors.some(function(candidateSelector) {
@@ -72,14 +74,16 @@ module.exports = function(el, options) {
 
       candidates = Array.prototype.slice.apply(candidates).concat(Array.prototype.slice.apply(slotCandidates));
       continue;
-    } else if (candidateIndex === 0) {
-      basicTabbables.push(candidate);
     } else {
-      orderedTabbables.push({
-        index: i,
-        tabIndex: candidateIndex,
-        node: candidate,
-      });
+      if (candidateIndex === 0) {
+        basicTabbables.push(candidate);
+      } else {
+        orderedTabbables.push({
+          index: i,
+          tabIndex: candidateIndex,
+          node: candidate,
+        });
+      }
     }
   }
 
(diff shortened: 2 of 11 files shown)


DRiFTy17:
I'll see if I can figure out why the tests failed, but not locally for me...

craigkovatch:
This seems like a significant change -- any idea about perf impact of writing your own querySelectorAll?

This might be something to switch on a flag rather than using for every call.

.eslintrc
```diff
@@ -2,7 +2,8 @@
   "env": {
     "node": true,
     "browser": true,
-    "mocha": true
+    "mocha": true,
+    "es6": true
```
davidtheclark: This config option should be added as-needed instead of globally, so ES6 is *blocked* for most files.

index.js
```diff
@@ -1,14 +1,15 @@
+
 module.exports = function(el, options) {
   options = options || {};
-
+
```
davidtheclark: Please clean up these added spaces.

index.js
```diff
@@ -17,29 +18,14 @@ module.exports = function(el, options) {
...
-    ) {
-      candidates = Array.prototype.slice.apply(candidates);
-      candidates.unshift(el);
-    }
-  }
+  var candidates = deepQuerySelectorAll(el, candidateSelectors, options.includeContainer);
 
   var candidate, candidateIndex;
-  for (var i = 0, l = candidates.length; i < l; i++) {
+  for (var i = 0; i < candidates.length; i++) {
```
davidtheclark: Please remove this unrelated and unnecessary change. The same goes for the space change below.

index.js
```diff
@@ -71,6 +57,55 @@ module.exports = function(el, options) {
...
 
+/**
+ * Walks the DOM tree starting at a root element and checks if any of its children 
+ * match the provided selectors. Similar to the native `querySelectorAll` except
+ * that it will traverse the shadow DOM as well as slotted nodes.
+ * @param {Element} rootElement The element to start querying from.
+ * @param {string[]} selectors An array of CSS selectors.
+ * @param {boolean} checkRootElement True if the provided root element is to be matched against the selectors.
+ */
+function deepQuerySelectorAll(rootElement, selectors, checkRootElement) {
```
davidtheclark: Is there an npm package that implements this? It doesn't seem right to me that this library, tabbable, would include its own implementation of such a function. Would be much better to use a shared and thoroughly tested public package.

index.js
```diff
@@ -82,7 +117,7 @@ function createIsUnavailable(elementDocument) {
   // "off" state, so we need to recursively check parents.
 
   function isOff(node, nodeComputedStyle) {
-    if (node === elementDocument.documentElement) return false;
+    if (node === elementDocument.documentElement || node instanceof ShadowRoot) return false;
```
davidtheclark: Won't this addition throw a `SyntaxError` for any browsers in which `ShadowRoot` is not defined?

craigkovatch:
This is probably my own ignorance, but what is the purpose of looking for tabbables in the shadow DOM? Can the user tab to something that's not in the regular DOM?

DRiFTy17:
@davidtheclark Thanks for taking a look at this, I fully expected there to be some discussion and iterations before this change was merged (if at all). You (and @craigkovatch) definitely bring up some good points!

I'll revert the small changes you noted, as well as come up with a different way to detect an instance of `ShadowRoot` as that is definitely an issue for the foreseeable future in browsers that don't support it... Thanks for pointing that one out.

As for your question about using an existing variation of this sort of deep querySelectorAll function. I totally agree that it doesn't make sense to have something like this live only inside tabbable. Due to web components still being relatively new, and not really widely adopted as of yet, there isn't too much out there aside from Polymer and a some various UI components. This sort of thing very well may have been written somewhere, but if so, I'm not aware of any packages that expose it. I did a fair bit of searching around for something when working on this actually. I think you'll find that because this stuff is still mostly in its infancy, there's a lot left to be desired. With the web component spec still changing, some of this stuff may end up being changed down the road again, but it's hard to say.

I think making this functionality configurable (off by default) is the way to go for now for tabbable as @craigkovatch mentioned, and it could be used if needed that way. 

Let me know how you'd like to proceed with that `deepQuerySelectorAll` function, and I can either do some more research on an existing solution, or move it out to it's own package.

craigkovatch:
@DRiFTy17 I would still love to understand the use case better, as I do a ton of work professionally with accessibility, keyboard and otherwise. Also, it sounds like it might be a better choice for you to write the infrastructure you need (e.g. DQSA) as a separate library. Then tabbable, if it makes sense, could import that library rather than baking it in.

Aaaaand while we're talking about infrastructure, @davidtheclark it would be great if tabbable could be packaged up as a module at some point. It's currently a little ugly to consume in e.g. TypeScript.

DRiFTy17:
@craigkovatch Thanks for the thoughts on using a configurable option for turning this on. I think that is the way to go for sure as that is likely going to be the use case 99% of the time for now at least until web components are more widely adopted.

I have not done any perf tests on the custom function. I'm sure it's slower, but that's definitely something I'll investigate to get an idea of how much performance we would sacrifice by using something like this.

As for your question about the purpose of this change to find tabbables in the shadow DOM. This comes down to the fact that yes, a user can tab through and interact with the UI no problem for elements in the shadow DOM, but when using JavaScript to find these elements, you must traverse the shadow root as if it's its own document. If an element is focused within a custom element, when asking for the `activeElement` within the main document, you will only receive the custom element as the active element. The shadow root for each custom element has its own `activeElement` that can be used to get the actual active element with that shadow DOM tree.

Because tabbable is based on the fact that it knows the tabbable elements, if an element within a shadow root is focused (or focusable) it currently does not pick up those elements so it does not allow them to be focused. Another issue is that using `<slot>` causes content projection. This needs to be taken into account to preserve the proper tab order even though that content lives somewhere different than where it's rendered on screen by the web component.

craigkovatch:
> If an element is focused within a custom element, when asking for the activeElement within the main document, you will only receive the custom element as the active element. The shadow root for each custom element has its own activeElement that can be used to get the actual active element with that shadow DOM tree.

Oh right because in web dev NOTHING CAN EVER BE EASY 😂

DRiFTy17:
@craigkovatch I agree, I am currently using tabbable (and focus-trap) within a TypeScript project and it could use some TypeScript support (I'm happy to do this work as well).

I am in the same boat. I take a lot of care in writing accessible websites and UI components. My main focus professionally is working on websites, UI components, and libraries (mostly internal). There are ADA requirements for everything that I write. I also can't tell you how many times I've had to rewrite the same UI components for various UI frameworks in recent years... This is the exact reason for moving to web components and using a standard to build these with vanilla JavaScript so it will work everywhere. I fully expect to see a big rise in using native web components soon, it's been picking up recently but there are certainly some drawbacks.

davidtheclark:
> it would be great if tabbable could be packaged up as a module at some point.

It is already, right? That's how I use consume — `import` or `require` it from a project that uses a module bundler. Do you mean something more specific?

@DRiFTy17 What do you think about 

- Packaging the shadow-dom-sensitive `querySelectorAll` into its own npm package.
- Using it conditionally in tabbable, behind a flag, which web component users can set to `true`.

??

DRiFTy17:
@davidtheclark I think by module he meant just having a TypeScript declaration file for it to make it easier to use within a TypeScript project. Importing/requiring it works fine as you mentioned.

I am on board with your suggestions for sure. I think that is the best way to go, and will not affect anyone who doesn't care about web components. I myself will use the flag for various elements within the same application because some of them use shadow DOM and others use light DOM.

When I get some time I'll create the npm package and update the PR. We can review what it looks like at that point and go from there. Thanks!

DRiFTy17:
@davidtheclark I've created a new npm module `shadow-utils` and consumed it within tabbable. Take a look and let me know what you think, and if there is anything else you can see that you'd like changed/added/removed.

Also, I added a tiny declaration file (there wasn't much to declare haha) so that should make it a little better for TypeScript. I can do the same within the PR for `focus-trap` as I already have those TypeScript interfaces written in a separate project.

craigkovatch:
> It is already, right? That's how I use consume — import or require it from a project that uses a module bundler. Do you mean something more specific?

tabbable when consumed in TypeScript requires us to "break the rules" with imports because it's packaged as a single global function:

```
// tslint:disable-next-line:no-require-imports
import tabbable = require('tabbable');
```

Typically we want to do this instead:
```
import { tabbable } from 'tabbable'
```

DRiFTy17:
@craigkovatch You can do that already like this: `import * as tabbable from 'tabbable';`

craigkovatch:
I feel bad that this thread has been hijacked but since we're here. Both of those syntaxes give me tsc errors:
```
// [ts] Module ''tabbable'' has no default export.
import tabbable from 'tabbable';

// [ts] Module ''tabbable'' resolves to a non-module entity and cannot be imported using this construct.
import * as tabbable from 'tabbable';
```
This might be our fault, maybe our typings file is Bad And We Should Feel Bad?:
```
declare function tabbable(element: HTMLElement): HTMLElement[];
declare module 'tabbable' {
  export = tabbable;
}
```

DRiFTy17:
@craigkovatch Ok, I see what's going on now. You do have your declaration file written correctly to match the state of the module. That was my mistake, I didn't realize that TypeScript adds a default export during compilation in certain cases until now.

I think what needs to happen is we need to choose whether we want a default export or not. I'm typically against using default exports because they are kind of restrictive and require a breaking change to import statements if removed. Likewise if we use only named exports, then the import syntax in this module is uglier as you mentioned. Tabbable seems like a library that will likely ever only export one thing, but would we want the flexibility to export other things later on? Probably.

Just so we're on the same page. If we choose a default export, then it might look similar to the following (feel free to correct me if I'm missing something):

**Update tabbable to have a default export:**
```
function tabbable(el, options) { ... }
function createIsUnavailable(elementDocument) { ... }

module.exports = tabbable;
module.exports.default = tabbable;
```

**Here is the declaration file to match:**
```
declare module 'tabbable' {
  function tabbable(element: Element, options?: ITabbableOptions): Element[];

  export interface ITabbableOptions {
    includeContainer?: boolean;
    deep?: boolean;
  }

  export default tabbable;
}
```

**Now the following import statements/usages will work:**
- `import tabbable from 'tabbable';`
    - `const tabbables = tabbable(el);`
- `import { default as tabbable } from 'tabbable';`
    - `const tabbables = tabbable(el);`
- `import tabbable, { ITabbableOptions } from 'tabbable';`
    - `const options: ITabbableOptions = { deep: true };`
       `const tabbables = tabbable(el, options);`
- `import * as tabbable from 'tabbable';`

**The following import statements now do not work properly:**
- `import { tabbable } from 'tabbable';`
    - `const tabbables = tabbable(el); // FAILS`
    - `const tabbables = tabbable.tabbable(el); // OK`
    - This would be considered a breaking change for anyone currently using the library this way, and would need to be updated to use: `import { default as tabbable } from 'tabbable;'`

So this kind of gets us the opposite of what we have now, but feels more clean at the same time. Thoughts?

On the other hand if we just use named exports (like it is currently), then no matter what, everyone will always import things the same way: `import { tabbable, ITabbableOptions, SomeFutureTabbableFeature } from 'tabbable';` which I tend to like doing myself.

DRiFTy17:
@davidtheclark Just curious if you have gotten a chance over the past few weeks to look this over again with the new changes in place?

Also, do you either you or @craigkovatch have any thoughts on the typings/exports that I mentioned in my last comment?

davidtheclark:
@DRiFTy17 I got thrown off the scent a little by the TypeScript discussion so this slipped off my radar. I'll try to look again soon. Let's move the TypeScript business to another issue and leave any of that out of this PR?

DRiFTy17:
@davidtheclark No problem, yeah this has gotten a little off topic for sure. I've updated the PR to remove any typings. Makes sense to add that as a separate PR.

davidtheclark:
@DRiFTy17: I've decided I'd like to keep tabbable as it is and suggest that you create a new shadow-dom-capable tabbable package. For now I don't want to disturb the stability of the project and add this bulk and complexity to it for what I'd consider to be a pretty rare edge case. Instead of using a flag to get shadow-dom access, the user can switch to a different package. I'd rather keep my maintenance as-minimal-as-possible here; and I think shadow-dom users might be better off with a package maintained by someone sharing their needs.

DRiFTy17:
@davidtheclark Sure thing, sounds good.

davidtheclark:
Ok, @DRiFTy17. Sorry again for all the back and forth, and great work on your end! If you create a package like this please let me know so we can add a link in this README.

PR #25, opened 2018-05-05
Fix tab order bug with SVGs in IE

### LINKED ISSUE #24: IE11, svg with tabindex=0 does not have tabIndex property
In IE11, a SVG with `tabindex="0"` does not have a `tabIndex` property so gets added to the `orderedTabbables` array no matter where it is in the DOM.

```js
candidateIndex = parseInt(candidate.getAttribute('tabindex'), 10) || candidate.tabIndex;
```

Always triggers the OR condition if `tabindex="0"`, which then assumes `tabIndex` is a valid property. SVGs return `undefined` for the `tabIndex` property.

davidtheclark:
Closes #24.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## HEAD
+
+- Fix bug causing SVG elements to precede elements they should follow in the tab order in IE.
+
 ## 1.1.2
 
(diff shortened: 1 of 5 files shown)


davidtheclark:
🤦‍♂️ thanks for pointing that out.

PR #32, opened 2018-08-02
simplify function to test for visibility

tinynow:
I'm not sure if there is a performance benefit from the current method for getting "untouchable" hidden elements, but this pull request uses less code and no loops to do it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,198 +1,150 @@
 var candidateSelectors = [
-  'input',
-  'select',
-  'textarea',
-  'a[href]',
-  'button',
-  '[tabindex]',
-  'audio[controls]',
-  'video[controls]',
-  '[contenteditable]:not([contenteditable="false"])',
-];
-var candidateSelector = candidateSelectors.join(',');
+    'input',
+    'select',
+    'textarea',
+    'a[href]',
+    'button',
+    '[tabindex]',
+    'audio[controls]',
+    'video[controls]',
+    '[contenteditable]:not([contenteditable="false"])',
+  ];
+  var candidateSelector = candidateSelectors.join(',');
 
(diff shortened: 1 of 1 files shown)


davidtheclark:
Hi @tinynow. I'm having a hard time reading this diff and understanding your suggestion. Do you think you could clean it up?

index.js
```diff
@@ -1,198 +1,150 @@
...
+    return !checked || checked === node;
+  }
 
-UntouchabilityChecker.prototype.isUntouchable = function isUntouchable(node) {
-  if (node === this.doc.documentElement) return false;
-  var computedStyle = this.doc.defaultView.getComputedStyle(node);
-  if (this.hasDisplayNone(node, computedStyle)) return true;
-  return computedStyle.visibility === 'hidden';
-}
+  function isHidden(node) {
```
tinynow: @davidtheclark This function replaces the UntouchabilityChecker.  All the other changes result from that.
offsetParent tests for display none: https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom



I don't know why the diff is so weird. This is my first pull request on github. Thank you very much for your work.

davidtheclark: Ohhh, interesting trick! 

I confirmed in Chrome, Firefox, and IE11. Have you confirmed in any other browsers?

It looks to me like you indented the entire file — I wonder if that's causing all the unnecessary diff.

kfranqueiro: Seems like it - https://github.com/davidtheclark/tabbable/pull/32/files?w=1 shows a much shorter diff.

stof:
I took this over at #37 to fix conflict with newer changes.

davidtheclark:
Thanks, @stof. Closing this PR in favor of #37.

PR #33, opened 2018-08-24
Fix `ReferenceError: Element is not defined` in Node.js

rgrove:
Since there's no `Element` global in Node.js, it's not safe to assume that `Element.prototype` exists. Looks like this broke in 3.0.0.

Tabbable isn't actually useful in server-side code, but universal React components may still end up loading Tabbable via focus-trap-react when rendering on the server.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,9 @@ var candidateSelectors = [
 ];
 var candidateSelector = candidateSelectors.join(',');
 
-var matches = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
+var matches = typeof Element === 'undefined'
+  ? function () {}
+  : Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -11,7 +11,9 @@ var candidateSelectors = [
 ];
 var candidateSelector = candidateSelectors.join(',');
 
-var matches = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
+var matches = typeof Element === 'undefined'
+  ? function () {}
+  : Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
```
davidtheclark: Do you think it's preferable to do this or to much `matches` within the scope of the `tabbable` function and add a `if (typeof window === 'undefined') return` at the top of that function's scope?

rgrove: I think that would be perfectly fine too. No preference here. 🙂

davidtheclark:
Should be released now in 3.1.1 👍

PR #35, opened 2018-12-11
Fix error when element is inside a Shadow DOM

### LINKED ISSUE #34: getComputedStyle fails with shadow dom
I'm trying to use focus-trap in a StencilJS component and the various checks that rely on getComputedStyle fail when it hits a shadow root.

Pretty sure this could be fixed with a simple check, I'll take a crack at a PR. Thanks!

```
7vfqzmxv.entry.js:formatted:2 Uncaught TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
    at p.hasDisplayNone (7vfqzmxv.entry.js:formatted:2)
    at p.hasDisplayNone (7vfqzmxv.entry.js:formatted:2)
    at p.hasDisplayNone (7vfqzmxv.entry.js:formatted:2)
    at p.hasDisplayNone (7vfqzmxv.entry.js:formatted:2)
    at p.isUntouchable (7vfqzmxv.entry.js:formatted:2)
    at l (7vfqzmxv.entry.js:formatted:2)
    at s (7vfqzmxv.entry.js:formatted:2)
    at i (7vfqzmxv.entry.js:formatted:2)
    at w (7vfqzmxv.entry.js:formatted:2)
    at Object.activate (7vfqzmxv.entry.js:formatted:2)
```

acdvorak:
In Web Components, the root element is a document fragment rather than an `<html>` element.

Fixes #34 and https://github.com/material-components/material-components-web/issues/4131

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,7 +167,7 @@ function UntouchabilityChecker(elementDocument) {
 // getComputedStyle accurately reflects `visibility: hidden` of ancestors
 // but not `display: none`, so we need to recursively check parents.
 UntouchabilityChecker.prototype.hasDisplayNone = function hasDisplayNone(node, nodeComputedStyle) {
-  if (node === this.doc.documentElement) return false;
+  if (node.nodeType !== Node.ELEMENT_NODE) return false;
 
     // Search for a cached result.
     var cached = find(this.cache, function(item) {
--- a/test/tabbable.test.js
+++ b/test/tabbable.test.js
@@ -270,6 +270,16 @@ describe('tabbable', function() {
         var n7 = assertionSet.getFixture('radio').getDocument().getElementById('radio-c');
         assert.ok(tabbable.isFocusable(n7));
       });
+
+      it('supports detached elements', function() {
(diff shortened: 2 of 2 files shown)


davidtheclark:
Released in 4.0.1.

kfranqueiro:
I don't think this fix has been included in any release? 4.0.1 is the latest `focus-trap` release, but it's depending on `tabbable` 3.1.1 which was released before this change.

davidtheclark:
Ugh, thanks @kfranqueiro. I think I tried to do this too quickly during my holiday break and ended up confusing myself. I'll fix it this morning.

davidtheclark:
Ok, unless I still have my head twisted around the wrong way, this should be fixed now: 3.1.2 of tabbable includes this change, and 4.0.3 of focus-trap updates to the new version of tabbable.

kfranqueiro:
Just tried installing them locally and it looks like the change is there now. Thanks!

PR #37, opened 2019-02-25
Optimize and simplify the function to test for visibility

stof:
This is taking over #32, fixing conflicts with newer changes.

This change is not only simplifying the code. It is also optimizing it a lot. When running focus-trap on an element containing a thousand checboxes (and a few other tabbable items), `updateTabbableNodes` (which is mostly a call to `tabbable`) went from 945ms to 15ms (on a rather powerful desktop browser).

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,11 +18,9 @@ var matches = typeof Element === 'undefined'
 function tabbable(el, options) {
   options = options || {};
 
-  var elementDocument = el.ownerDocument || el;
   var regularTabbables = [];
   var orderedTabbables = [];
 
-  var untouchabilityChecker = new UntouchabilityChecker(elementDocument);
   var candidates = el.querySelectorAll(candidateSelector);
 
   if (options.includeContainer) {
@@ -36,7 +34,7 @@ function tabbable(el, options) {
   for (i = 0; i < candidates.length; i++) {
     candidate = candidates[i];
 
-    if (!isNodeMatchingSelectorTabbable(candidate, untouchabilityChecker)) continue;
+    if (!isNodeMatchingSelectorTabbable(candidate)) continue;
 
     candidateTabindex = getTabindex(candidate);
     if (candidateTabindex === 0) {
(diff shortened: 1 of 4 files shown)


stof:
For now, this breaks the test `supports detached elements`.

This one was added in #35 when fixing support for using tabbable inside a shadow root. But I think the test itself is wrong: an element inside a shadow root is not the same than a detached element (an element inside a shadow root can be visible or hidden. A detached element is never visible as it is not even in the DOM)

stof:
note that the text title properly describes what the test is asserting. But I'm not sure that this case is something that tabbable needs to support (while I agree that support shadow roots makes sense).

stof:
OK, I managed to write a test confirming that shadow roots still work with the rewritten implementation (even though detached elements don't anymore).

stof:
And, test is green on Travis once it runs on a recent Firefox version (ShadowDom is enabled only in Firefox 63+ while Travis uses Firefox 56 by default)

index.js
```diff
@@ -154,47 +144,8 @@ function isTabbableRadio(node) {
...
-    return result;
-}
-
-UntouchabilityChecker.prototype.isUntouchable = function isUntouchable(node) {
-  if (node === this.doc.documentElement) return false;
-  var computedStyle = this.doc.defaultView.getComputedStyle(node);
-  if (this.hasDisplayNone(node, computedStyle)) return true;
-  return computedStyle.visibility === 'hidden';
+function isHidden(node) {
+  return node.offsetParent === null || getComputedStyle(node).visibility === 'hidden';
```
davidtheclark: Can you please add a comment above this line about why both conditions are needed and what they do?

stof:
@davidtheclark the test added in #35 was *not* covering the issue reported in the bug. The bug was talking about using tabbable inside a shadow root. The test was trying to use it in a DocumentFragment detached from the DOM (where things cannot be visible, as they are not in the DOM at all yet).
The new test I added is actually the one covering the reported issue.

stof:
@davidtheclark comment added

kfranqueiro:
It looks like the OP of the issue on MDC Web deleted their glitch instance, so I can't test with the original problem case. But for whatever it's worth, I took this branch's tests and tested against 3.1.1 and 3.1.2, and it fails on both of those versions with a similar error to what was originally reported, but works against the branch. So if anything, this seems like it fixes the shadow DOM issue more appropriately than Andy's original PR, as far as I can tell.

davidtheclark:
Thanks @kfranqueiro and @stof. Let's release this, then.

PR #38, opened 2019-03-05
Setup rollup build, add TS types

Andarist:
I'd actually want to tweak build in https://github.com/davidtheclark/focus-trap-react but thought I could work on the whole thing bottom-up and that's how I ended up doing this.

If you like it I will prepare followup PRs in related packages.

⚠️ This is a breaking change due to exports shape change. The proposed form makes this library way more tree-shakeable (static properties are treeshaking/DCE enemies)


davidtheclark:
Hi @Andarist. I'm not interested in overhauling the codebase this dramatically. Is there a problem you're aiming to solve here?

Andarist:
Basically, I wanted to optimize https://github.com/davidtheclark/focus-trap-react (bundle size), but decided to work on this bottom-up. This PR makes this lib more bundler friendly, more future proof (ESM modules, CJS only as fallback). This also provides TS typings.

I understand this might seem like a lot but in its essence, this just split a single file into multiple to improve readability and changes exports shape - it exports things separately instead of attaching utils as static properties of the main export.

If you are going to vote against this then I would be ok with it, no harsh feelings - I've decided to work on this knowing that this might get rejected. You can tell me which parts do you like - i.e. updating deps, eslint, using babel, whatever and I can remove unwanted parts to get some of the things merged in.

That being said - I strongly think that this improves the package overall, maybe not dramatically, but every bit counts 😉

davidtheclark:
@Andarist If you'd like to contribute a change, please open an issue describing what you'd like to do, and please open separate issues for separate changes (e.g. an issue for using Babel, an issue for TS typings). This would help focus any time spent on discussing the changes. 

Non-breaking version upgrades to dependencies are always welcome as PRs.

brian-lagerman:
I put in a PR for ES6 Support on focus-trap davidtheclark/focus-trap#84 without having seen this. However, this would further lighten the load as tabbable wouldn't need to be bundled anymore from focus-trap's perspective. Is there any progress on the concerns with this PR?

PR #51, opened 2020-03-31
Add UMD build

bbenjamin:
I'm evaluating replacements for jQuery UI's `:tabbable` in Drupal core, since jQueryUI is end-of-life. I have a strong preference for tabbable, but to be used in Drupal it would needs to be available as a global variable definition. The UMD files generated by this patch work great in Drupal, so I'm hoping you're open to adding this as an option. 

Also happy to assist on other issues that the next release is waiting on.


davidtheclark:
@bbenjamin I'm not actively developing this project, so it would be great if you'd like to take over maintenance (I can add you as a collaborator) or fork it!

PR #52, opened 2020-04-12
Fix isHidden function for elements with position: fixed

shamsartem:


shamsartem:
I have a button with position: fixed that closes a dialog. It works fine on Firefox for reasons described here (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) but on chrome it doesn't work. I hope with my fix it will at least work consistently. Hope I am being useful. Thanks for the great library!

stefcameron:
@shamsartem I realize this is a long time coming, but if you're still keen on fixing this, I'd take it in if you also added a test to check that old behavior isn't changed, but is indeed fixed: https://github.com/focus-trap/tabbable/tree/master/test

stefcameron:
(I just took over as a new maintainer)

shamsartem:
@stefcameron Thanks for the response and for maintaining this valuable project. I didn't add any tests because it seems like there are already enough tests for that, for example this one checks exactly this behavior: https://github.com/focus-trap/tabbable/blob/e2ac54c3853a57fb73cae40c8ba8fe66f043e6fe/test/fixtures/jqueryui.html#L62
If you think something else is needed please tell me exactly what it is and I will do it. I really hope this is merged because it's still an issue on my website. Thanks

stefcameron:
> @stefcameron Thanks for the response and for maintaining this valuable project.

My pleasure! 😄 

> I didn't add any tests because it seems like there are already enough tests for that, for example this one checks exactly this behavior:
> 
> https://github.com/focus-trap/tabbable/blob/e2ac54c3853a57fb73cae40c8ba8fe66f043e6fe/test/fixtures/jqueryui.html#L62

I see. Thanks for pointing that out. I'm still learning how tests here work. I took over for `focus-trap` and `focus-trap-react` too, so lots to get familiar with...

> If you think something else is needed please tell me exactly what it is and I will do it.

So if I understand this bug correctly, it's that if you have:
```html
<div style="position: fixed">
  <button/>
</div>
```
`tabbable()` will not find it in Chrome (but will in FF) with the current code, but with your fix, it'll find it. Is that right? If so, then this is what I feel is missing in the tests: The part of the jqueryui fixture you pointed out doesn't appear to test the "fixed" case, only that `<input id="displayNoneAncestor-input" />` isn't considered tabbable because it's in a hidden container. We should add a div with position "fixed" that contains a button that is now expected to be found.

> I really hope this is merged because it's still an issue on my website. Thanks

I'm keen to take it in! Just need to sort out the test.

shamsartem:
Ok I added a test which shows that using .offsetParent leads to incorrect behavior. Hope it's fine now. Thanks

shamsartem:
I also noticed that ChromeHeadless is now used for tests but karma-firefox-launcher is still in package.json but not used. I suggest using both Firefox and Chrome for tests to avoid such inconsistencies between browsers like this one in the future. Thanks

stefcameron:
> Ok I added a test which shows that using .offsetParent leads to incorrect behavior. Hope it's fine now. Thanks

Thank you, looks good!

> I also noticed that ChromeHeadless is now used for tests but karma-firefox-launcher is still in package.json but not used. I suggest using both Firefox and Chrome for tests to avoid such inconsistencies between browsers like this one in the future. Thanks

I'm happy to take that in. Can you add `FirefoxHeadless` to the Karma `browsers` list?

shamsartem:
Ok I added FirefoxHeadless

stefcameron (approved):
Thank you for this fix! I really appreciate your help and contribution.

PR #53, opened 2020-06-02
Change public exports shape

### LINKED ISSUE #39: Make this package tree-shakeable
I'm building a UI library which exports a bunch of stuff, only some of my components rely on this package so it would be desirable that if a consumer of my package doesn't use component relying on `tabbable` it would be tree shaken out of the final bundle. However it's not the case right now.

Let's focus on 2 common tree-shaking deopts:
- no ESM bundle (this package) exports only CJS format
- top-level static properties, as in `module.exports = tabbable; tabbable.isTabbable = isTabbable;`

Action items:
- migrate source to ESM + build CJS format from it, point to CJS from package.json#main and to ESM from package.json#module so bundlers can pick appropriate file
- change exports shape to `export default tabbable; export { isTabbable, isFocusable };`

Intent to implement - **yes** (was already implemented in https://github.com/davidtheclark/tabbable/pull/38 )

Andarist:
**Why?**

Improving tree-shakeability of this package. Usage of static properties like this usually retains the containing object (in this case a function) even if it stays unused. This matters when `tabbable` is being used as transitive dep.

Partially fixes #39 

⚠️ this is a breaking change

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Dependencies: _none_.
 ### tabbable
 
 ```js
-import tabbable from 'tabbable';
+import { tabbable } from 'tabbable';
 
 tabbable(rootNode, [options]);
 ```
--- a/README.md
+++ b/README.md
@@ -47,14 +47,14 @@ npm install tabbable
 
 Dependencies: _none_.
 
-You'll need to be compiling CommonJS with your bundler of choice.
-
 ## API
 
 ### tabbable
 
-```
-tabbable(rootNode, [options])
+```js
+import tabbable from 'tabbable';
+
+tabbable(rootNode, [options]);
 ```
 
 Returns an array of ordered tabbable node within the `rootNode`.
@@ -76,18 +76,22 @@ Type: `boolean`. Default: `false`.
 
 If set to `true`, `rootNode` will be included in the returned tabbable node array, if `rootNode` is tabbable.
 
-### tabbable.isTabbable
+### isTabbable
 
-```
-tabbable.isTabbable(node)
+```js
+import { isTabbable } from 'tabbable';
+
+isTabbable(node);
 ```
 
 Returns a boolean indicating whether the provided node is considered tabbable.
 
-### tabbable.isFocusable
+### isFocusable
 
-```
-tabbable.isFocusable(node)
+```js
+import { isFocusable } from 'tabbable';
+
+isFocusable(node);
 ```
 
 Returns a boolean indicating whether the provided node is considered _focusable_.
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -36,7 +36,6 @@ const cjs = [
       format: 'cjs',
       esModule: false,
       sourcemap: true,
-      exports: 'named',
     },
   },
   {
@@ -46,7 +45,6 @@ const cjs = [
       format: 'cjs',
       esModule: false,
       sourcemap: true,
-      exports: 'named',
     },
     plugins: [...commonPlugins, terser({ ...terserOptions, toplevel: true })],
   },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -36,6 +36,7 @@ const cjs = [
       format: 'cjs',
       esModule: false,
       sourcemap: true,
+      exports: 'named',
     },
   },
   {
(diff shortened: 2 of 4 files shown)


Andarist:
> Also, what, in your opinion, is still missing in this to fully address #39?

I would have to review this, can do that later this week.

> Since we bundle as CJS also, I'd like to avoid potential bundling issues and not have a default export. Could we just export all functions as export { tabbable, isTabbable, isFocusable }?

Sure, you can go with named exports only route.

Andarist:
Ok, I've adjusted the PR according to your comments - please take a look.

> Also, what, in your opinion, is still missing in this to fully address #39?

1. `/* #__PURE__ */` comment before a call [here](https://github.com/focus-trap/tabbable/blob/81ccf3668ac3332cd16300cbf9bb6709abcf81e8/src/index.js#L13)
2. tools don't want to remove top-level getters because of the potential side-effects and, unfortunately, there is no perfect way to make those treeshakeable. There are only quirk solutions for it - so it depends if you care about having [this](https://github.com/focus-trap/tabbable/blob/81ccf3668ac3332cd16300cbf9bb6709abcf81e8/src/index.js#L15) tree-shakeable. If yes, then I could give some guidance on how to fix this statement.

stefcameron:
> 1. `/* #__PURE__ */` comment before a call [here](https://github.com/focus-trap/tabbable/blob/81ccf3668ac3332cd16300cbf9bb6709abcf81e8/src/index.js#L13)

I don't understand this one. Because of the `.join()` call?

> 2. tools don't want to remove top-level getters because of the potential side-effects and, unfortunately, there is no perfect way to make those treeshakeable. There are only quirk solutions for it - so it depends if you care about having [this](https://github.com/focus-trap/tabbable/blob/81ccf3668ac3332cd16300cbf9bb6709abcf81e8/src/index.js#L15) tree-shakeable. If yes, then I could give some guidance on how to fix this statement.

I'm not seeing how `matches` could be removed. It seems to be used by all of `tabbable`, `isTabbable`, and `isFocusable`. What am I missing?

Andarist:
> This is looking good, thanks for updating to only named exports. What about also marking the module as side-effect-free with sideEffects: false in package.json? Wouldn't that also help (when using bundlers like Webpack that support it)?

Only to some extent - this only helps for ignoring imported files and as you only distribute a single file, it doesn't bring that much value. I would have to recheck but maybe it would help in such a case:
```js
// foo.js
import { tabbable } from 'tabbable'

export function A() {
  tabbable()
}

export function B() {
  console.log()
}

// entry.js
import { B } from './foo'
console.log(B)
```
but it wouldn't help cases like this:
```js
// entry.js
import { isTabbable } from 'tabbable'
console.log(isTabbable)
```
If anything gets required from `tababble` then its single file gets included and everything in it as well. Later this goes through tree-shaking algorithm (I don't treat `"sideEffects"` as part of this) and unused stuff might get removed, or not. It depends on how the code is written and if the optimizer is able to assess that unused stuff don't create any side-effects that could alter the behavior of the program if they would have been removed.

> I don't understand this one. Because of the .join() call?

Yes. Tools don't know what `.join` does and it could potentially even make an important request so they have to bail out from removing this call and leave this expression in the consuming bundles. This comment is a hint for a minifier to drop it **if** the resulting value stays unused.

> I'm not seeing how matches could be removed. It seems to be used by all of tabbable, isTabbable, and isFocusable. What am I missing?

Like in the first scenario that I have included in this post. It's about transitive dependencies. Tree-shaking (in my opinion) is not purely about scenarios when your library gets actually used but also about scenarios when it stays unused. Like when a `lib` library depends on `tabbable`, but the user only imports from `lib` stuff that does not depend on `tabbable`. Ideally, no `tabbable` code should get included in the final bundle.

stefcameron:
@Andarist Thanks for taking the time to answer my 3 questions, that's very helpful. This will be good to help with #39.

I do think that it would be worth adding `sideEffects: false`. I've seen it work well in certain scenarios where I've dealt with a library that was getting pulled in simply because of Webpack not knowing it was safe to exclude when nothing was using it anymore (i.e. the example you gave).

Might as well add the `PURE` comment also, as you [suggested](https://github.com/focus-trap/tabbable/pull/53#issuecomment-679110885) earlier.

PR #60, opened 2020-08-27
Add `prepare` script

Andarist:
1. It will run before publishing a package so u wont forget about doing a fresh build then running `npm publish`
2. it will also allow installing this package from GitHub (as in - npm will fetch devDeps and build this package when installing this like `npm install tabbable@focus-trap/tabbable`)


stefcameron:
@Andarist  Thanks for thinking of this! With #58, I'm addressing (1) with `prepublishOnly` that will run the build. I'm wondering about (2). When we publish the package, we include the built source in `./dist`. Why do we need to worry about building it again on install? Installing from GitHub is new to me. I've never done that, so perhaps there's some technical thing I'm not aware of.

Andarist:
(2) This is for the case when installing from the remote git repository - in which you don't keep `./dist`. In this case npm will install dev deps and run your build script to actually prepare `./dist` so the package becomes usable. This doesn't affect regular installs from npm. This is useful because one can install from any hash/branch/whatever of this repo - without waiting for a maintainer to publish to npm.

stefcameron:
🤔  According to https://docs.npmjs.com/misc/scripts, `prepare` is run not only before the package is published, but also on `npm install` -- which would be unnecessary in this case because we publish `./dist`. I feel like installing straight from the remote repo instead of npm is not a normal use case (I have no data to back this up, but I'll bet most installs are from npm, not GitHub). So I see `prepare` as adding an unnecessary build step on every install via npm for the sake of convenience for very few installs straight from the repo. If you're installing from the repo, you can easily run the `build` script we provide to fix this, and since it's common not to commit built, distributed files, it shouldn't be too surprising that installing from the repo doesn't provide the built files out-of-the-box.

Now that we have `prepublishOnly` in place, I'm including not to add `prepare`. Makes sense?

Andarist:
According to those docs - `prepare` is only run after *local* npm install. This means that if I clone this repo and run `npm install` then this script will be executed. This **wont** be executed when installing this package from npm, but it **will** be executed when installing this package from GitHub.

I don't find this to be problematic that this is also executed after local npm installs - how much time one really lose by it? And how often do you run `npm install` in this repo?

> If you're installing from the repo, you can easily run the build script we provide to fix this

In such a case I need to write a custom logic and educate myself which script should I execute - I also need to install all of your devDeps, ideally with the very same versions that you are specifying in your package.json manifest. This is not trivial task.

> since it's common not to commit built, distributed files, it shouldn't be too surprising that installing from the repo doesn't provide the built files out-of-the-box.

Well, as there is a sanctioned way of solving this (prepare script) without committing dist files (which I also don't like) then I don't see why we wouldn't like to support this out-of-the-box behavior.

> for the sake of convenience for very few installs straight from the repo

YMMV, this surely is a way less common way to install packages but I found it very useful in the past, on multiple occasions. It allows me to skip the whole fork/publish dance for simple changes that I can benefit from before the maintainer decides to publish the package to the npm.

stefcameron:
@Andarist Thanks for your patience in helping me understand this. I finally managed to do some local testing with a local NPM registry.

I created this package.json
```json
{
  "name": "prepare-pkg",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "prepare": "echo \"Running prepare script\""
  },
  "author": "",
  "license": "ISC"
}
```
I published it, and then installed it. I observed:
1. `prepare` was run before publishing -- this is good.
2. `prepare` was not run on `npm install prepare-pkg --registry localhost:port`. That's good for me.

Case 2 surprised me, though. I still find the NPM docs super not clear on this `prepare` script. I tried installing the package from a local directory, `npm install ../prepare-pkg`, prepare script still was not run. Why not? This is as local as it gets, the docs on the script don't explicitly say it only runs when installing from a local git repo, do they? I don't see that anywhere.

So when *exactly* is this script run, other than just before publishing?

My main concern is that it ends-up being run when someone does `npm install tabbable` -- run unnecessarily because this way, they will get the published `./dist` folder with the already-built output. There's no need for an extra build step in that case.

I guess I've sort of validated that this won't be the case, but I still don't feel ready to take this in because I don't know what `npm install` use case it actually gets executed from.

(I could see applying this to focus-trap and focus-trap-react also.)

Andarist:
> So when exactly is this script run, other than just before publishing?

1. when you run `npm install` in a project that has a `prepare` script
2. before publishing
3. when installing a package from a git repo (this has been undocumented for quite some time, I've learned about it from the release notes)

> My main concern is that it ends-up being run when someone does npm install tabbable -- run unnecessarily because this way, they will get the published ./dist folder with the already-built output. There's no need for an extra build step in that case.

It won't be run in this case - it's explicitly designed to not run in this situation.

stefcameron (approved):
@Andarist Thanks for those details! That finally makes it nice and clear for me. 😄  Merging... (and I'll do the same in `focus-trap` and `focus-trap-react` too!)

PR #68, opened 2020-09-04
Mark module as having no side effects

### LINKED ISSUE #39: Make this package tree-shakeable

stefcameron:
Finally fixes #39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,11 @@
   - In case a form parent element exists, include only nested radio inputs from that form.
   - Ignore checked radio elements from forms different from the one the validated node belongs to.
   - NOTE: This may result in *less* radio elements being flagged as tabbable depending on context from the "root" node given to `tabbable()`.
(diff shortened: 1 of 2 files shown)


stefcameron:
Changeset not needed for this. PR has update to CHANGELOG.

PR #70, opened 2020-09-04
Tree shaking improvements

Andarist:
Related to: https://github.com/focus-trap/tabbable/issues/39

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "5.0.0",
   "description": "Returns an array of all tabbable DOM nodes within a containing node.",
   "main": "dist/index.min.js",
-  "module": "dist/index.esm.min.js",
+  "module": "dist/index.esm.js",
   "types": "index.d.ts",
   "sideEffects": false,
   "files": [
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "tabbable",
   "version": "5.0.0",
   "description": "Returns an array of all tabbable DOM nodes within a containing node.",
-  "main": "dist/index.min.js",
+  "main": "dist/index.js",
   "module": "dist/index.esm.js",
   "types": "index.d.ts",
   "sideEffects": false,
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -61,16 +61,6 @@ const esm = [
       sourcemap: true,
     },
   },
-  {
-    input,
-    output: {
-      file: 'dist/index.esm.min.js',
-      format: 'esm',
-      esModule: true,
-      sourcemap: true,
-    },
-    plugins: [...commonPlugins, terser(terserOptions)],
-  },
 ];
 
 const umd = [
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -61,6 +61,16 @@ const esm = [
       sourcemap: true,
     },
   },
+  {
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -91,7 +91,9 @@ function isNodeMatchingSelectorFocusable(node) {
   return true;
 }
 
-let focusableCandidateSelector = candidateSelectors.concat('iframe').join(',');
+let focusableCandidateSelector = /* #__PURE__ */ candidateSelectors
```
Andarist, line 94: I've missed this one previously.

rollup.config.js
```diff
@@ -61,16 +61,6 @@ const esm = [
       sourcemap: true,
     },
   },
-  {
```
Andarist, line 64: By minifying the `"module"` you drop comments so existing `/* #__PURE__ */` become useless (they disappear). 

I don't see a point in pre-minifying the "module" export. It's only going to be used in a combination with a bundler and bundlers are usually already configured to minify code for production - so by minifying upfront, you only make the content of a file more obscure and you discourage trying to debug issues related to it. I for instance very often dive into `node_modules` to debug various issues and being able to actually read the source code (well, code close to source code) and being able to modify it to test stuff is very helpful.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "5.0.0",
   "description": "Returns an array of all tabbable DOM nodes within a containing node.",
   "main": "dist/index.min.js",
```
stefcameron, line 5: Shouldn't we also not reference the minified version of the CJS bundle for the same reasons?

rollup.config.js
stefcameron: That's a good point about those comments getting dropped! I was trying to figure out how to prevent those specific comments from being removed from the minified bundle but couldn't.

Also good point about not pointing to the minified bundle.

stefcameron: I was trying over in https://github.com/focus-trap/tabbable/pull/72/files#diff-ff6e5f22a9c7e66987b19c0199636480R15 (though comment retention wasn't the point of that change, but still, I was thinking of this too; any way, moot point with changing the package's main/module to point to unminified bundles anyway, that retain all the comments).

package.json
stefcameron: I think we should reference the non-minified version of the CJS bundle for the same reasons.

rollup.config.js
stefcameron: Still, we've already published the minified version of the ESM bundle, so I think we should keep this in. Now that we're pointing `module` to the non-minified version, that will work for bundlers. If someone wants the minified version for some reason, then they have it. If we remove the minified build now that it's been published, that feels like a breaking change to me, which would require a new major, and I really don't want to do another major release for some time now.

Andarist:
Both changes are OK for me - is there anything else you'd like to do/discuss here or is this one ready to be merged?

stefcameron (approved):

stefcameron:
> Both changes are OK for me - is there anything else you'd like to do/discuss here or is this one ready to be merged?

Cool! Nothing more to do. I was just giving you time to check over the changes I made to make sure we were both on the same page. 😄 

I'll merge it shortly.

PR #41, opened 2019-04-03
Add focusable function to API

vonagam:
Add a function to select all focusable elements (not only tabbable ones).

Is there any objections to adding such functionality?

If no, i will proceed to adding simple tests.

PR #73, opened 2020-09-05
Add tests for focusable()

stefcameron:
API was added in #41.


stefcameron:
No changeset required. This is only tests.

PR #74, opened 2020-09-06
Add focusable() type definition

vonagam:


stefcameron (approved):
Thanks for remembering the typings!

PR #84, opened 2020-09-17
The TypeScript return type of `tabbable` has been changed from `Array<Element>` to `Array<HTMLElement | SVGElement>`

Andarist:
Please verify this but I believe that the change is correct - it seems that the raw `Element` interface doesn't have `focus()` method on it. That method comes from `HTMLOrSVGElement` which both `HTMLElement` and `SVGElement` extend.

###  Features and Bug Fixes

- [x] Typings added/updated.

### DIFF
--- a/.changeset/blue-lions-wink.md
+++ b/.changeset/blue-lions-wink.md
@@ -1,5 +1,5 @@
 ---
-'tabbable': minor
+'tabbable': patch
 ---
 
-The TypeScript return type of `tabbable` has been changed from `Array<Element>` to `Array<HTMLElement | SVGElement>`.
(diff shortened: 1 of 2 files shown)


stefcameron (approved):
Thanks for this! Definitely, `Element` was not right since it doesn't provide the `focus()` method. `HTMLElement` definitely provides it. [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement) is tricky because it doesn't state any methods, only what it inherits from. I learned that `SVGElement` gets its `focus()` method thanks to implementing the [HTMLOrForeignElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement) interface. Learn something every day!

stefcameron:
I changed this from a minor to patch since I see this more as a bug fix than a new feature.

Andarist:
> I changed this from a minor to patch since I see this more as a bug fix

Beauty of changesets 🦋 ❤️

Issue #99, opened 2020-10-07
version 5.1.1 contains an arrow function


khamiltonuk:
I am trying to use the library in internet explorer 11 

in the transpiled version there is an arrow function 
I can see it coming from line 49 
```js
let tabbableNodes = orderedTabbables
    .sort(sortOrderedTabbables)
    .map((a) => a.node)
    .concat(regularTabbables);
```
https://github.com/focus-trap/tabbable/blob/master/src/index.js#L49

We have two options, we can use a traditional function expression in the `index.js`, or we can configure the babel settings to have a target of internet explorer 9 to match the version support in the docs

stefcameron:
@khamiltonuk Thanks for filing this issue! It ended-up being brought-up over on https://github.com/focus-trap/focus-trap/pull/173 and we had a discussion there.

Bringing that over to here now:

---

Webpack can be configured to work around this. The issue is you're wanting to use ESM as a modern module system, but still expecting to have old browser-compatible code inside those modules, and not wanting to transpile any third-party modules. Is that correct (just trying to understand the needs)?

It feels strange to have diverging targets in the same bundle: ES modules (which is technically ES6) and ES5 JS. But I admit I see the same issue in my bundles that reference `focus-trap` if I set my Babel target to `ie: "11"` (I end-up with the same issue with the arrow function @khamiltonuk pointed out earlier).

Looks like this is a problem for a lot of libraries, and Webpack is keeping its position on prioritizing `module` over `main` for 3 years now: https://github.com/webpack/webpack/issues/5756 (still open)

A lot of packages seem to have solved this by adding a `browser` entry to their `package.json` (because Webpack's [default order](https://webpack.js.org/configuration/resolve/#resolvemainfields) is `mainFields: ['browser', 'module', 'main']`), which is something we could do with `focus-trap` and `tabbable`.

Apparently, [this is the browser field spec](https://github.com/defunctzombie/package-browser-field-spec), but it doesn't really get into specifics about the nature of the referenced files.

Still, given its name "browser", it feels like it should point to the UMD build.

Thoughts?

---

I just tried this locally with tabbable's package.json having "browser: dist/index.umd.js" (with Webpack) and the resulting bundle that includes tabbable look nice. Webpack took the right UMD "path" for CJS and plopped the result right into the bundle:
```javascript
}, function(module, exports, __webpack_require__) {
  !function(exports) {
    "use strict";
[...]

PR #100, opened 2020-10-07
fix: use a traditional function expression for old browsers

khamiltonuk:
Issue https://github.com/focus-trap/tabbable/issues/99

An arrow function is causing syntax errors in ie11

This is not the best solution but it's the fastest, please close if you don't like it,
a better long term solution would be to make sure that babel is correctly configured 
I have made a pull request here https://github.com/focus-trap/tabbable/pull/103

###  Features and Bug Fixes

- [X] Issue being fixed is referenced.
~~[ ] Test coverage added/updated.~~
~~[ ] Typings added/updated.~~
~~[ ] README updated (API changes, instructions, etc.).~~
~~[ ] Changes to dependencies explained.~~
- [X] Changeset added (run `yarn changeset` locally to add one, follow prompts).

### DIFF
--- a/.changeset/lemon-birds-watch.md
+++ b/.changeset/lemon-birds-watch.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+fix: remove arrow function and use tradional function expression for older browser support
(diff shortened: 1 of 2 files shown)


stefcameron:
Thanks for this quick fix, but I don't think it's the right approach. Closing per my comments on #99.

PR #103, opened 2020-10-07
fix: remove arrow function causing issues for old browsers using a babel

khamiltonuk:
Issue https://github.com/focus-trap/tabbable/issues/99

An arrow function is causing syntax errors in ie11
I have added a babel plugin to transform arrow functions

###  Features and Bug Fixes

~~[ ] Test coverage added/updated.~~
~~[ ] Typings added/updated.~~
~~[ ] README updated (API changes, instructions, etc.).~~

### DIFF
--- a/.changeset/afraid-trees-brush.md
+++ b/.changeset/afraid-trees-brush.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+fix: remove arrow function causing issues for old browsers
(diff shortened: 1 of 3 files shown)


stefcameron (changes requested):
Thanks for this update to Babel. If we were to handle this with Babel, then we should simply set the target to IE 11 in the `@babel/presets` configuration in babel.config.js instead of using this transform. That should work.

That said, it's diverging targets for the same bundle: ES modules (ES6) and ES5 JS.

I think the right fix is to just add `browser: ./dist/index.umd.js` to the package.json.

Please see my comments on #99.

stefcameron:
Again, thank you for quickly submitting a PR to help fix this! I'm closing this in favor of #106

PR #106, opened 2020-10-07
Transpile the ESM bundle to ES5 but still use ES module system

### LINKED ISSUE #99: version 5.1.1 contains an arrow function

stefcameron:
Fixes #99

See https://github.com/webpack/webpack/issues/5756

The initial approach was to add a `browser` target to package.json and
point it at the UMD build so that bundlers would pick that up instead of
the ESM build, since they (except for Rollup) tend to prefer `browser`
first, then `module` (ESM), then `main`.

But that would potentially affect tree-shaking, and we don't want to
hamper that optimizing process.

So this PR transpiles the code in the ESM bundle down to ES5, so that
it runs in IE9+ like the CJS and UMD bundles do, while keeping the
import/export ESM statements.

###  Features and Bug Fixes

- [x] ~Test coverage added/updated.~
- [x] ~Typings added/updated.~
- [x] ~README updated (API changes, instructions, etc.).~
- [x] ~Changes to dependencies explained.~

### DIFF
--- a/.changeset/giant-grapes-nail.md
+++ b/.changeset/giant-grapes-nail.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+Fix #99: Add `browser` target in package.json for use by bundlers. Both Webpack and Rollup can use this target above `module` and `main`, when specified, and our UMD bundle is transpiled down to basic ES5 for maximum browser compatibility. Our ESM bundle targets ESM browsers (which means is does not support IE 11). Webpack does this [by default](https://webpack.js.org/configuration/resolve/#resolvemainfields). Rollup, however, needs to be [configured](https://github.com/rollup/plugins/tree/master/packages/node-resolve#mainfields).
--- a/.changeset/giant-grapes-nail.md
+++ b/.changeset/giant-grapes-nail.md
@@ -2,4 +2,4 @@
 'tabbable': patch
 ---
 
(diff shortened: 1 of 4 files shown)


stefcameron:
@khamiltonuk @bparish628 Trying to fix tabbable... I think this should get things fixed for the bundlers.

bparish628:
This would work in the sense that the code used by webpack would use transpiled code, but the main reason people like esm is because it gets treeshaken out.

I was looking at how other libraries (chakra-ui, emotion) do it and it seems that the code is transpiled to the current browser supported by the library in all of the supported module formats. It would be nice if this library also did that, where all formats would be transpiled down to the browser that this library says it supports (ie9). 

I think most projects use esm so things get treeshaken out (with webpack), not necessarily on what browser's can run that code (that is what babel is for).

I think browser support and **bundler** module definitions are two different things. Browser support is babel  and import/exports are the bundler.

This is your library, so don't want to tell you what to do! Just giving my 2 cents 🙂

stefcameron:
@bparish628 Thanks for your thoughts. I appreciate it! Sure seems like conflicting intents for ESM bundles, but you make a good point about tree shaking with ESM imports and exports, and the `browser` approach would remove that possibility. I'll digest this a bit...

stefcameron:
I thought about this for a bit. Tree shaking is definitely compelling, and we've done work in this module recently to make it tree-shakable (see #39), so I wouldn't want to thwart that effort. I think I went astray when I equated the _module system_ with the _browser target_, as you (@bparish628) pointed out:

> I think browser support and bundler module definitions are two different things. Browser support is babel and import/exports definitions are the bundler's job.

When I think of ESM as just the module system, then it makes sense to transpile the code inside to the same browser target as the CJS and UMD builds.

So that's what I will do! Hopefully that will unbreak things happily. 😄

PR #114, opened 2020-10-18
Details and summary element fixes

idoros:
This PR fixes some issues with the details/summary elements:

1. ignore elements nested under a closed details element with the exception of the first summary element.
2. ignore any summary elements except the first.
3. pickup details elements as tabbable in case they don't have any summary.

###  Features and Bug Fixes

- ~Issue being fixed is referenced.~
- [x] Test coverage added/updated.
- ~Typings added/updated.~
- ~Changes to dependencies explained.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ The following are considered tabbable:
 - `<textarea>` elements
 - `<a>` elements with `href` or `xlink:href` attributes
 - `<audio>` and `<videos>` elements with `controls` attributes
-- `<summary>` element directly under a `<details>` element
+- the first `<summary>` element directly under a `<details>` element
 - elements with the `[contenteditable]` attribute
 - anything with a non-negative `tabindex` attribute
 
@@ -25,6 +25,7 @@ Any of the above will _not_ be considered tabbable, though, if any of the follow
 - has a negative `tabindex` attribute
 - has a `disabled` attribute
 - either the node itself _or an ancestor of it_ is hidden via `display: none` or `visibility: hidden`
+- is nested under a closed `<detailed>` element (with the exception of the first `<summary>` element)
 - is an `<input type="radio">` element and a different radio in its group is `checked`
 
 **If you think a node should be included in your array of tabbables _but it's not_, all you need to do is add `tabindex="0"` to deliberately include it.** (Or if it is in your array but you don't want it, you can add `tabindex="-1"` to deliberately exclude it.) This will also result in more consistent cross-browser behavior. For information about why your special node might _not_ be included, see ["More details"](#more-details), below.
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ The following are considered tabbable:
 - `<a>` elements with `href` or `xlink:href` attributes
 - `<audio>` and `<videos>` elements with `controls` attributes
 - the first `<summary>` element directly under a `<details>` element
+- `<details>` element without a `<summary>` element
 - elements with the `[contenteditable]` attribute
 - anything with a non-negative `tabindex` attribute
 
@@ -25,7 +26,7 @@ Any of the above will _not_ be considered tabbable, though, if any of the follow
 - has a negative `tabindex` attribute
 - has a `disabled` attribute
 - either the node itself _or an ancestor of it_ is hidden via `display: none` or `visibility: hidden`
-- is nested under a closed `<detailed>` element (with the exception of the first `<summary>` element)
+- is nested under a closed `<details>` element (with the exception of the first `<summary>` element)
 - is an `<input type="radio">` element and a different radio in its group is `checked`
 
 **If you think a node should be included in your array of tabbables _but it's not_, all you need to do is add `tabindex="0"` to deliberately include it.** (Or if it is in your array but you don't want it, you can add `tabindex="-1"` to deliberately exclude it.) This will also result in more consistent cross-browser behavior. For information about why your special node might _not_ be included, see ["More details"](#more-details), below.
--- a/src/index.js
+++ b/src/index.js
@@ -193,6 +193,12 @@ function isTabbableRadio(node) {
 function isHidden(node) {
   if (getComputedStyle(node).visibility === 'hidden') return true;
 
+  const isDirectSummary = node.matches('details>summary:first-of-type');
+  const nodeUnderDetails = isDirectSummary ? node.parentElement : node;
+  if (nodeUnderDetails.matches('details:not([open]) *')) {
+    return true;
+  }
+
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -25,6 +25,7 @@ Any of the above will _not_ be considered tabbable, though, if any of the follow
 - has a negative `tabindex` attribute
 - has a `disabled` attribute
 - either the node itself _or an ancestor of it_ is hidden via `display: none` or `visibility: hidden`
+- is nested under a closed `<detailed>` element (with the exception of the first `<summary>` element)
```
stefcameron, line 28: `<detailed>` -> `<details>`

src/index.js
```diff
@@ -193,6 +193,12 @@ function isTabbableRadio(node) {
 function isHidden(node) {
   if (getComputedStyle(node).visibility === 'hidden') return true;
 
+  const isDirectSummary = node.matches('details>summary:first-of-type');
+  const nodeUnderDetails = isDirectSummary ? node.parentElement : node;
+  if (nodeUnderDetails.matches('details:not([open]) *')) {
+    return true;
+  }
```
stefcameron, line 200: Would it make sense to move this into an IF block?
```javascript
  if (node.matches('details>summary:first-of-type')) {
    return node.parentElement.matches('details:not([open]) *');
  }
```

In any case, based on samples in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details, it seems `<details>` elements on their own, even if not open, are tabbable, whether a `<summary>` exists or not inside of it.

My impression is we aren't considering that case here. WDYT?

It seems it's actually the `<details>` element that's tabbable, not the `<summary>`. The later just affects that gets displayed in the toggle, whether it's just "Details" (default without summary) or it's some other text (specified in the summary)...

test/fixtures/details.html
```diff
@@ -1,8 +1,11 @@
 <details>
-  <summary id="details-a-summery">details A title</summary>
+  <summary id="details-a-summary">details A title</summary>
```
stefcameron, line 2: Thanks for fixing that typo! 🙂

stefcameron (commented):
@idoros Thank you for taking the initiative to post this PR to fix a bug! Much appreciated. Some thoughts in comments...

src/index.js
idoros: > Would it make sense to move this into an IF block?

It would have to be 2 IF blocks with duplicate selector
```
if (
    node.matches('details>summary:first-of-type') &&
    node.parentElement.matches('details:not([open]) *')
  ) {
    return true;
  } else if (node.matches('details:not([open]) *')) {
    return true;
  }
```

> `<details>` elements on their own, even if not open, are tabbable

Good catch! From what I can see the actual `activeElement` for a details element with no summary is the details element itself, while the summary element get the focus if it exist (calling `focus()` on the details element will move the focus to the summary element).

I added details elements with no summary to be picked up as tabbable.

stefcameron: Thanks for the JSBin, very helpful! Cool, so we found something more to fix! 🎉 

About the IF statement, though, I'm still wondering if we could optimize this. It would be nice if we didn't have to test every node that gets passed to this function as being a summary element, and then the other query.

Could we not do this instead?

```javascript
function isHidden(node) {
  if (getComputedStyle(node).visibility === 'hidden') return true;

  // 👉 isn't this sufficient to see if it's a node that hidden under a collapsed <details> element?
  if (node.matches('details:not([open]) *')) {
    return true;
  }

  while (node) {
    if (getComputedStyle(node).display === 'none') return true;
    node = node.parentElement;
  }

  return false;
}
```

Otherwise, it's a `<details>` element itself, and if it's hidden, then the first check would get it, or the first check in the `while` loop below. Or am I missing something subtle?

src/index.js
```diff
@@ -97,7 +98,12 @@ function isTabbable(node) {
 }
 
 function isNodeMatchingSelectorFocusable(node) {
-  if (node.disabled || isHiddenInput(node) || isHidden(node)) {
+  if (
+    node.disabled ||
+    isHiddenInput(node) ||
+    isHidden(node) ||
+    isDetailsWithSummary(node)
```
stefcameron, line 105: I think upon reading this for the first time, this will seem counter-intuitive to a reader who isn't familiar with the nuance of details with/out summary inside. Could you add a comment here about the fact that when it contains a summary, it's the summary that's focusable, not the details? (Even I had to ponder about this one just now!)

src/index.js
```diff
@@ -129,13 +135,15 @@ function getTabindex(node) {
...
   //  `tabIndex` of -1 when the 'tabindex' attribute isn't specified in the DOM,
   //  yet they are still part of the regular tab order; in FF, they get a default
   //  `tabIndex` of 0; since Chrome still puts those elements in the regular tab
-  //  order, consider their tab index to be 0
+  //  order, consider their tab index to be 0.
   if (
-    (node.nodeName === 'AUDIO' || node.nodeName === 'VIDEO') &&
+    (node.nodeName === 'AUDIO' ||
+      node.nodeName === 'VIDEO' ||
+      node.nodeName === 'DETAILS') &&
```
stefcameron, line 146: Nice catch! I see that the `<summary>` element gets `tabIndex === 0` so we're good there too with line 129 above.

src/index.js
idoros: This checks 2 options

- `direct first summary`
    1. check that it's actually is the first summary of a details element (which will make it focusable even when it's parent details is not open)
    2. check that the **summary's details (it's parent)** isn't deeply nested in a closed details element
- `any other element` - just check that the element isn't deeply nested in a closed details element

The code you suggested would find a `direct first summary` in a closed details as hidden:
```
if (node.matches('details:not([open]) *')) {
    return true;
 }
```

I could replace the first check with an imperative check (check type, check parent, check previous siblings), but I'm not sure it would be faster, and it would definitely be more code.

I could maybe make it a tiny bit faster by just checking the first condition (type), but again, that probably is just an insignificant performance optimization, and maybe even not...
```
const isDirectSummary = node.tagName === 'SUMMARY' && node.matches('details>summary:first-of-type');
// instead of
const isDirectSummary = node.matches('details>summary:first-of-type');
```

stefcameron: Ah, OK. That's the detail I was missing in my head: the nesting. Thank you for being patient with me on this thread as I worked things out!

stefcameron (approved):
Looks good!

stefcameron:
@all-contributors add @idoros for bug, code

stefcameron:
FYI, this was published in 5.1.3.

PR #169, opened 2020-12-08
Escape CSS selector special characters in radio button names

### LINKED ISSUE #168: isTabbableRadio doesn't properly escape query selectors
It's possible for radio button names to contain characters that are invalid when used in a CSS selector (e.g. `"`).  Below is an example that causes the `tabbable` function to throw an error due to an invalid selector constructed at: <https://github.com/focus-trap/tabbable/blob/cd06b988e9b1924405ba406c1a503046af2687a2/src/index.js#L105>

**HTML:**
```html
<div>
  <input type="radio" id="huey" name="&quot;duck&quot;" value="huey"
         checked>
  <label for="huey">Huey</label>
</div>

<div>
  <input type="radio" id="dewey" name="&quot;duck&quot;" value="dewey">
  <label for="dewey">Dewey</label>
</div>
```

**JS:**
```js
import { tabbable } from 'tabbable';
// Throws DOMException: Document.querySelectorAll: 'input[type="radio"][name=""duck""]' is not a valid selector
tabbable(document.body);
```

It's also possible to inject values into the selector with a name like:

```html
<input type="radio" id="dewey" name="&quot;] .some-class[data-attr=&quot;some-attribute" value="dewey">
```

A potential solution is [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape), but it does require a polyfill for IE.

tidychips:
Fixes #168.  Uses `CSS.escape` to escape special characters in the radio button `name` attribute when constructing the selector here:

<https://github.com/focus-trap/tabbable/blob/cd06b988e9b1924405ba406c1a503046af2687a2/src/index.js#L105>

Also adds a caveat to the documentation about requiring a polyfill for `CSS.escape` when running on IE.

###  Features and Bug Fixes

- [x] Test coverage added/updated.
- [ ] ~Typings added/updated.~
- [ ] ~Changes to dependencies explained.~

### DIFF
--- a/.changeset/soft-days-invite.md
+++ b/.changeset/soft-days-invite.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+fix crash when radio button name attributes contain CSS selector special characters
--- a/.changeset/soft-days-invite.md
+++ b/.changeset/soft-days-invite.md
@@ -2,4 +2,4 @@
 'tabbable': patch
 ---
 
-fix crash when radio button name attributes contain CSS selector special characters
+fix crash when radio button name attributes contain CSS selector special characters (#168)
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Basically IE9+.
 
 Why? It uses [Element.querySelectorAll()](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll) and [Window.getComputedStyle()](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle).
 
-**Note:** When used with any version of IE, [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) needs [polyfilled](https://www.npmjs.com/package/css.escape) for tabbable to work properly with radio buttons that have `name` attributes containing special characters.
+**Note:** When used with any version of IE, [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) needs a [polyfill](https://www.npmjs.com/package/css.escape) for tabbable to work properly with radio buttons that have `name` attributes containing special characters.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Basically IE9+.
 
 Why? It uses [Element.querySelectorAll()](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll) and [Window.getComputedStyle()](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle).
 
+**Note:** When used with any version of IE, [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) needs [polyfilled](https://www.npmjs.com/package/css.escape) for tabbable to work properly with radio buttons that have `name` attributes containing special characters.
+
 ## Installation
 
 ```
--- a/src/index.js
+++ b/src/index.js
@@ -96,14 +96,13 @@ const getCheckedRadio = function (nodes, form) {
   }
 };
 
-const escapeCSS =
-  typeof window === 'undefined' ||
-  typeof window.CSS === 'undefined' ||
-  typeof window.CSS.escape !== 'function'
-    ? function (string) {
-        return string;
-      }
-    : window.CSS.escape;
+const escapeCSS = function (string) {
+  return typeof window === 'undefined' ||
+    typeof window.CSS === 'undefined' ||
+    typeof window.CSS.escape !== 'function'
+    ? string
+    : window.CSS.escape(string);
+};
 
 const isTabbableRadio = function (node) {
   if (!node.name) {
--- a/src/index.js
+++ b/src/index.js
@@ -100,8 +100,8 @@ const escapeCSS =
   typeof window === 'undefined' ||
   typeof window.CSS === 'undefined' ||
   typeof window.CSS.escape !== 'function'
-    ? function (value) {
-        return typeof value === 'string' ? value.replace(/"/g, '\\"') : value;
+    ? function (string) {
+        return string;
       }
     : window.CSS.escape;
 
--- a/src/index.js
+++ b/src/index.js
@@ -96,31 +96,36 @@ const getCheckedRadio = function (nodes, form) {
   }
 };
 
-const escapeCSS = function (string) {
-  return typeof window === 'undefined' ||
-    typeof window.CSS === 'undefined' ||
-    typeof window.CSS.escape !== 'function'
-    ? string
-    : window.CSS.escape(string);
-};
-
 const isTabbableRadio = function (node) {
   if (!node.name) {
     return true;
   }
   const radioScope = node.form || node.ownerDocument;
+
+  const queryRadios = function (name) {
+    return radioScope.querySelectorAll(
+      'input[type="radio"][name="' + name + '"]'
+    );
+  };
+
   let radioSet;
-  try {
-    radioSet = radioScope.querySelectorAll(
-      'input[type="radio"][name="' + escapeCSS(node.name) + '"]'
-    );
-  } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error(
-      'Looks like you have a radio button with a name attribute containing invalid CSS selector characters and need the CSS.escape polyfill: %s',
-      err.message
-    );
-    return false;
+  if (
+    typeof window !== 'undefined' &&
+    typeof window.CSS !== 'undefined' &&
+    typeof window.CSS.escape === 'function'
+  ) {
+    radioSet = queryRadios(window.CSS.escape(node.name));
+  } else {
+    try {
+      radioSet = queryRadios(node.name);
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -44,6 +44,8 @@ Basically IE9+.
 
 Why? It uses [Element.querySelectorAll()](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll) and [Window.getComputedStyle()](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle).
 
+**Note:** When used with any version of IE, [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) needs [polyfilled](https://www.npmjs.com/package/css.escape) for tabbable to work properly with radio buttons that have `name` attributes containing special characters.
```
stefcameron, line 47: "...needs [polyfilled]..." -> "...needs a [polyfill]..."

src/index.js
```diff
@@ -96,13 +96,20 @@ const getCheckedRadio = function (nodes, form) {
   }
 };
 
+const escapeCSS =
+  typeof CSS === 'undefined' || typeof CSS.escape !== 'function'
+    ? function (string) {
+        return string;
```
stefcameron, line 102: Let's at least plug the injection hole either way by doing `return string.replace(/"/g, '\\"')` and add a test for both the case where `CSS` is defined and where it isn't, if possible.

src/index.js
```diff
@@ -96,13 +96,20 @@ const getCheckedRadio = function (nodes, form) {
   }
 };
 
+const escapeCSS =
+  typeof CSS === 'undefined' || typeof CSS.escape !== 'function'
```
stefcameron, line 100: To make sure this doesn't trip-up any folks doing server-side rendering, is it better to test `typeof window !== 'undefined' && (typeof window.CSS === 'undefined' || typeof window.CSS.escape !== 'function')`?

I realize anything declared on `window` is technically a global and doesn't need `window.` to access it, but I don't have that much experience with server-side rendering to know all the ins and outs, but enough to have been tripped-up myself with missing `typeof window !== 'undefined'` in a few places before to be weary about it...

.changeset/soft-days-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+fix crash when radio button name attributes contain CSS selector special characters
```
stefcameron, line 5: Please add the related issue at the end: `(#168)`

stefcameron (changes requested):
@tidychips Thank you for being proactive and posting a PR for #168 -- HUGE help!

src/index.js
tidychips: I'll add a replace, but for mitigating injections specifically, I think `\` also needs added for the characters to replace (As a raw name of `\"` would still defeat it).  There may also be ways to defeat that too, my experience with these sort of issues has always been to defer to library given how notorious they are to handle incorrectly yourself.  Though, as noted in the issue, the consequences for getting it wrong or not handling it 100% are pretty minor compared to most injection scenarios.  

I'll add a commit to replace `"`, and then defer to what you think is reasonable in terms of the other special characters to cover for this partial polyfill.

src/index.js
tidychips: I added a check for `typeof window`, but don't really have any SSR experience myself.  I'll do some research and see if there's a better way to handle it.

src/index.js
stefcameron: This is true. I fiddled a little more with it and it's a downward slope toward, well, `CSS.escape`. And reflecting a little more on it, the escape is still within the realms of `querySelectorAll()`, which should itself protect against worse things than getting an invalid selector, or one that causes other elements to be selected.

I guess I'm more convinced now to just to nothing and return whatever `string` is if there's no`CSS.escape` available (what you had in the first place). For IE users (hopefully very _few_ now...) that also don't have the polyfill, they'll get the same behavior as today (error), but now they'll have the option to polyfill and we'll support it.

Thanks for being willing to add (and adding) the check for quotes, but let's revert that.

idoros: Another option is to provide a safe 'quarySelectorAll' similar to 'matchas' in order to catch the possible error in browsers without 'CCS.escape'

stefcameron: Yes, I thought of that too, but then I got thinking that if we did that, or just wrapped in try..catch here, it would fail silently and be worse than the exception being thrown since users wouldn't know why the radios aren't being found. I guess we could try..catch and print a more informative error to the console in that case to be more helpful.

idoros: I think that for this specific flow I would rather get the different behavior then an error that might break some higher level functionality

stefcameron: So you'd rather we catch the error, maybe `console.error("Looks like you have a radio button with a name attribute containing invalid CSS selector characters and need the CSS.escape polyfill: %s", err.message)` and then skip the radio buttons and continue execution?

tidychips: I've added the `try / catch` and reverted the `replace` in the fallback.  I'd like to add a test but I'm having a hard time figuring out how to fit it in with the existing test structure.  The ways I've thought of to have the fallback be active in the test is either:

1. Add the IE karma plugin and run it (which is a whole can of worms)
1. Remove `escape` from the `CSS` object and re-import the module (hacky)
1. Change the logic to check if a fallback is necessary every time the function runs, instead of when the module is imported, and use mocks or environment variables to force the fallback in tests (not ideal as the implementation has test-specific logic, and it's not obvious what can cause a test failure if the implementation changes)
1. Use a test outside of a browser that can more easily setup mocks to force the fallback (not ideal to have one or two tests break from convention.  Also additional devops overhead and maintenance)

It's also very possible that I'm missing something obvious, so let me know any suggestions or ideas you might have for this @stefcameron @idoros

stefcameron: Could you just do something like this?
```javascript
it('should work with fallback', () => {
  const previous = CSS.escape;
  CSS.escape = undefined;

  // test...

  if (previous) {
    CSS.escape = previous;
  }
});
```
Maybe not the most elegant, but it's a pattern that has worked for me in the past in similar situations since `CSS` isn't a module that gets loaded (i.e. it doesn't need to be mocked in some special Karma way. It's just provided by the environment).

tidychips: Hmm, the issue I found with that approach is the `escapeCSS` function is only assigned when the module is imported.  So even if `CSS.escape` is set to undefined, all of the `tabbable` functions will still use the previous, valid, reference to `CSS.escape`.  It's possible to modify it to something like this:

```js
function escapeCSS(string) {
  return typeof window === 'undefined' ||
    typeof window.CSS === 'undefined' ||
    typeof window.CSS.escape !== 'function'
    ? string
    : window.CSS.escape(string);
}
```

but that comes with the drawback that it's very easy to update the function back to a form where the reference to the `escape` function used isn't always using the latest `CSS.escape` value.  If that happens, the test has no way of knowing and will continue to pass even though the fallback will no longer be tested.  There's no easy way with the current setup to detect that lack of coverage and fail.

tidychips: Well, I think I found something that works by writing assertions against the added `console.error` message.  Let me know if you think it looks okay.  Particularly with adding `sinon` as a devDependency to help out with stubbing, since it's pretty heavyweight compared to what the test needs it for.

stefcameron (approved):
Nice work thinking outside the box with `sinon`! I don't have any objections. It's just a dev library, not an actual package dependency that everyone will have to install.

And thanks for adding the `console.error()` notice, that should bring the necessary amount of attention in a more helpful manner than just allowing the exception to be thrown.

src/index.js
```diff
@@ -96,14 +96,34 @@ const getCheckedRadio = function (nodes, form) {
...
 const isTabbableRadio = function (node) {
   if (!node.name) {
     return true;
   }
   const radioScope = node.form || node.ownerDocument;
-  const radioSet = radioScope.querySelectorAll(
-    'input[type="radio"][name="' + node.name + '"]'
-  );
+  let radioSet;
+  try {
```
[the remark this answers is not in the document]
stefcameron: I don't understand what you mean by "safeQuerySelectorAll", and don't follow on the detriment of try..catch.

A modern browser will have CSS.escape and things should just work in that case. Otherwise, they fail, we catch, show a helpful message instead of crashing, and consider the radio not tabbable.

Are you concerned about some performance overhead?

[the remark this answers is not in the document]
idoros: In modern browsers escape exist and there is no reason to wrap with try/catch. My concern is for performance, although tbh I'm not sure if try/catch still has any penalty like it use to have.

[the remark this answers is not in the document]
stefcameron: I see. Maybe it did in older browsers, but I'm not concerned for modern ones, and also, if there were any overhead, it would only be on radio buttons. Otherwise, I don't know what else we could do except give up on being nice with the friendly error and just let it crash and trust people on older browsers will read the README note we've added. That behavior would be no different to how it's been for a very long time. But I like the helpful error, personally.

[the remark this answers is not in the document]
stefcameron: OK, I reversed the logic kind of. I think this is what you were getting at. We only `try..catch` if `CSS.escape` is not available. Otherwise, we assume there won't be an exception.

stefcameron:
@all-contributors add @tidychips for code

src/index.js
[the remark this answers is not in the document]
stefcameron: I'm moving forward with this.

Issue #115, opened 2020-10-20
getComputedStyled() causes a forced reflow


stefcameron:
https://github.com/focus-trap/tabbable/blob/master/src/index.js#L193-L202

In concert with https://github.com/focus-trap/focus-trap/issues/121

stefcameron:
For those, I wonder if we could get away with just looking at the style attribute instead of using the function.

ju1ius:
Hi, looking at the code I think the issue might be calling `getComputedStyle` on all ancestors [here](https://github.com/focus-trap/tabbable/blob/ec601812f8c1cf31d9c99561fb36bf389c50ae20/src/index.js#L154).

A workaround could be to check if the element's bounding client rect has a zero size, e.g:
```js
const hasZeroSize = (node) => {
  const {width, height} = node.getBoundingClientRect()
  return width === 0 && height === 0
}
const isVisuallyHidden = node => getComputedStyle(node).visibility === 'hidden'
const isHidden = node => (
  isFirstDirectSummaryInClosedDetails(node)
  || hasZeroSize(node)
  || isVisuallyHidden(node)
)
```

This would at least prevent style / layout recalc on the whole document...

idoros:
This change is a breaking change and in my opinion there are cases where focusable elements have zero size, like an accessibility link to content that is not visible until focused (double-tab at the beginning of [bbc.com](https://www.bbc.com/) for example).

`getComputedStyle` should be lazy and shouldn't cause reflow without accessing any layout properties, so the performance issue is not directly related to `getComputedStyle`, but to the traversal and check up the DOM.

ju1ius:
> This change is a breaking change

Indeed, although the tests pass.

> In my opinion there are cases where focusable elements have zero size, like skip to main content link for accessibility that is not visible until focused

In this particular case, no. Having a zero-size element with focusable content is considered a strong accessibility anti-pattern.
That's why most if not all websites implements this using a 1px by 1px element which is then clipped and/or moved out of the document w/ absolute positioning, i.e:
```css
.visually-hidden {
  overflow: hidden;
  position: absolute;
  height: 1px;
  width: 1px;
  padding: 0;
  border: 0;
  clip: rect(0, 0, 0, 0);
}
```

> getComputedStyle should be lazy and shouldn't cause reflow without accessing any layout properties

[...]

PR #185, opened 2020-12-22
avoid calling `getComputedStyle` on all ancestors of a candidate

ju1ius:
`getBoundingClientRect()` returns a zero size when a node or any of it's ancestor has it's `display` style property set to `none`.
We leverage that fact to avoid checking computed styles for all ancestors of a focusable candidate.

While this doesn't prevent all unwanted style / layout recalculations, this should help mitigate the issue in #115 and [focus-trap#121](https://github.com/focus-trap/focus-trap/issues/121)

<details>
<br/>

- Prop-types added/updated.
- Typings added/updated.
  - EXCEPTION: A Changeset is not required if the change does not affect any of the source files that produce the package bundle. For example, tooling changes, test updates, or a new dev-only dependency to run tests more efficiently should not have a Changeset since it will not affect package consumers.

</details>

### DIFF
--- a/.changeset/shiny-experts-exercise.md
+++ b/.changeset/shiny-experts-exercise.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+avoid calling `getComputedStyle` on all ancestors of a focusable candidate
--- a/src/index.js
+++ b/src/index.js
@@ -88,6 +88,12 @@ const isDetailsWithSummary = function (node) {
   return r;
 };
 
+const isFirstDirectSummaryInsideClosedDetails = function (node) {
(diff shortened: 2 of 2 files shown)


stefcameron:
@ju1ius Thank you for this PR, much appreciated.

stefcameron:
To @idoros's point, it does feel like we should probably be more cautious and put this behind a flag on the `options` object as opt-in behavior, keeping the current behavior as default, and see how the rest of the user base likes it.

Also, is `getBoundingClientRect` compatible with all the browsers this library tries to support? ... Checked: Looks like it for the `width` and `height` properties.

stefcameron (changes requested):
Please see my previous comment ^^^ 🙂

Issue #225, opened 2021-02-16
Use @testing-library/dom and Jest for tests


thawkin3:
@stefcameron Still interested in re-writing the test suite with DOM Testing Library and Jest? (https://testing-library.com/docs/dom-testing-library/intro/)

I was going to jump on this soon if you still think it'd be a good idea. Let me know!

stefcameron:
Hey, yes, for sure! Please go ahead if you have the time. 😀

thawkin3:
Sweet! Working on it now. 😄

stefcameron:
🎉 Thank you!

PR #226, opened 2021-02-17
Feature/dom testing library

thawkin3:
I've re-written the test suite using Jest and DOM Testing Library. 🎉 

Screen Shot 2021-02-17 at 8 23 53 PM

This PR is somewhat large, so the following summary of changes should help:

1. Added needed devDependencies: `@testing-library/dom`, `@testing-library/jest-dom`, `babel-jest`, `jest`, and `jest-watch-typeahead`
2. Added Jest configuration: `jest.config.js` file, `setupTests.js` file, ESLint configuration, Babel configuration, script in `package.json`, and GitHub workflow configuration
3. Wrote the unit tests in four separate files, one for each exported method: `focusable`, `isFocusable`, `tabbable`, and `isTabbable`
4. Updated README to mention the Jest unit tests
5. Updated README to no longer mention support for `a[xlink:href]` attributes
6. Stripped out all Karma-related code: test file, devDependencies, npm script, CI config, Karma config file, README info

Closes https://github.com/focus-trap/tabbable/issues/225

<details>
<br/>

- [x] Source changes maintain stated browser compatibility.
- [x] Unit test coverage added/updated.
- [x] E2E test coverage added/updated.
- ~Typings added/updated.~
- [x] Changeset added (run `yarn changeset` locally to add one, and follow the prompts).

</details>


test/tabbable.jest.test.js
```diff
@@ -0,0 +1,301 @@
...
+      const container = document.createElement('div');
+      container.innerHTML = svg;
+
+      const tabbableElements = tabbable(container);
+
+      expect(getTabbableIds(tabbableElements)).toEqual(expectedTabbableIds);
+    });
+
+    // TODO: The results we get here do not match the results we get in the legacy Karma test
+    it.skip('correctly identifies tabbable elements in the "radio" example', () => {
```
thawkin3, line 175: I need to look into this radio button test and why I was getting different results from the original Karma test.

test/tabbable.jest.test.js
```diff
@@ -0,0 +1,301 @@
...
+      const container = document.createElement('div');
+      container.innerHTML = radio;
+
+      const tabbableElements = tabbable(container);
+
+      expect(getTabbableIds(tabbableElements)).toEqual(expectedTabbableIds);
+    });
+
+    // TODO: The results we get here do not match the results we get in the legacy Karma test
+    it.skip('correctly identifies tabbable elements in the "radio" example without the `CSS.escape` functionality', () => {
```
thawkin3, line 195: Same here. I need to look into this radio button test and why I was getting different results from the original Karma test.

test/isFocusable.jest.test.js
```diff
@@ -0,0 +1,283 @@
...
+        '<a href="https://github.com/focus-trap/tabbable">Focusable</a>';
+
+      expect(isFocusable(getByText(container, 'Focusable'))).toBe(true);
+    });
+
+    // The README says `isFocusable` should return `true for:
+    // <a> elements with href or xlink:href attributes
+    // But this is returning `false`.
+    // Should the README be updated to not include `xlink:href` here?
+    it.skip('returns true for an `a` anchor element with an `xlink:href` attribute', () => {
```
thawkin3, line 52: I need to confirm this behavior manually in a real app, outside of a test suite.

test/isTabbable.jest.test.js
```diff
@@ -0,0 +1,278 @@
...
+        '<a href="https://github.com/focus-trap/tabbable">Tabbable</a>';
+
+      expect(isTabbable(getByText(container, 'Tabbable'))).toBe(true);
+    });
+
+    // The README says `isTabbable` should return `true for:
+    // <a> elements with href or xlink:href attributes
+    // But this is returning `false`.
+    // Should the README be updated to not include `xlink:href` here?
+    it.skip('returns true for an `a` anchor element with an `xlink:href` attribute', () => {
```
thawkin3, line 52: Same here. I need to confirm this behavior manually in a real app, outside of a test suite.

thawkin3:
@stefcameron Here's my first stab at it! I still have three TODO items that I mentioned in the PR description. The biggest thing that I wanted to confirm with you is that there's no reason to keep the Karma setup around anymore, right? If that's the case, there is a whole bunch of config code that I can strip out.

test/focusable.jest.test.js
```diff
@@ -0,0 +1,275 @@
...
+  basic,
+  nested,
+  jqueryui,
+  'non-linear': nonLinear,
+  'changing-content': changingContent,
+  svg,
+  radio,
+  details,
+  'shadow-dom': shadowDom,
+} = require('./fixtures/index.js');
```
stefcameron, line 12: For clarity in the test code, could we import this as `const fix = require(...)` and then have, for example, `fix.jqueryui` in the code when we're using a fixture? I think that would help with readability. Same thing for `test/tabbable.jest.test.js` where we also import the fixtures module.

test/isFocusable.jest.test.js
stefcameron: Like @idoros suggests, I think adding `a[xlink:href]` to the `candidateSelectors` list in the code would solve this. We should fix the code and leave the docs as they are (although it would've been like this for a _long_ time now, so if anyone was dependent on `<a xlink:href>` being tabbable, we should've heard by now, unless they created their own fork and fixed it there while the repo was unmaintained).

stefcameron (commented):
Great work, @thawkin3 , this is looking nice! Thanks for the nice summary in the PR description. I don't see any reason to keep Karma around. There's no other purpose for it beside running tests that are now run by Jest, so it can be ripped out rather than preserved.

test/focusable.jest.test.js
thawkin3: Will do!

thawkin3:
@stefcameron Awesome, I'll rip out Karma in another commit soon and add it to this PR. This PR is going to get a whole lot bigger with all the deletions!

test/isFocusable.jest.test.js
thawkin3: Thanks @stefcameron and @idoros, mostly I was wondering how you guys would like to tackle it. To me, it seems like the package currently _does not_ consider `a[xlink:href]` to be tabbable, like you said.

I'd probably recommend removing the line from the README first as part of this PR, and then maybe in a followup PR add the new line of code in the `candidateSelectors` array and put that line back in the README. That way the PR stays nice and focused on the test suite as opposed to also adding in new functionality to support `a[xlink:href]`.

You good with that approach? I'm fine either way since it's a small change, it just seems like a better approach to me to keep the PR focused to a single thing.

stefcameron: Well, looking at the Blame for `index.js` where the selectors are defined, it hasn't changed in over 3 years except 4 months ago when we added support for `<details>` elements. So IF `<a xlink:href>` had ever been supported at some point, it stopped being supported over 3 years ago. Going back to 5-6 years ago in the Blame, I still don't see `xlink` anywhere in `index.js`.

So I'm thinking it's safe to remove it from the README, but now we should include a Changeset with this change just to point out that we've removed adjusted the README to technically remove a feature, but that the code never actually supported it. Just for transparency. Technically, it would be breaking and so a major, but since it looks like it was never supported, perhaps a minor bump would be OK. I'd like to avoid a major bump in any case.

In light of all this history, I guess fixing the bug could be just as potentially problematic -- suddenly all these apps that for some reason use `<a xlink:href>` and `tabbable` now suddenly have all these new tabbable nodes and it messes everything up. Could be bad that way too.

I'm leaning on the side of removing from the README and adding a Changeset for a minor bump with a "heads-up" note.

thawkin3: Awesome, that sounds like a good plan to me. I'll remove that line from the README and add a changeset for this MR as a minor bump.

thawkin3: Actually now that I'm working on this, just to clarify, this would be a `patch` version as opposed to a `minor` version. Because this would be more like a "fix" than a new "feature" being added.

thawkin3:
> In regard to the radio button tests, the difference you get from the previous test code is that the new code doesn't attach the fixtures to the DOM. It doesn't break the behavior of most cases, but radio buttons have some query usage to find other group elements. And it make sense that "tabbable" elements would be attached to the DOM.

@idoros Well I learned something new today! Thanks for pointing that out. To avoid any other possible surprises, I made sure to actually append all of the test examples to the DOM now. You were spot on, and that fixed the radio button tests.

thawkin3:
@stefcameron @idoros This behemoth of a PR is ready for a second round of review! Here's a quick summary of what I've changed since our last conversation:

- Stripped out Karma and everything related to it
- Removed the info about the `xlink:href` attribute in the README
- Fixed the skipped Jest tests
- Updated the Jest tests to actually append the container content to the document body
- Added a `patch` changeset

test/isFocusable.jest.test.js
stefcameron: Yeah, I was debating between `patch` and `minor`. My thinking was that since `minor` is for new functionality, and we're _removing_ functionality (that hasn't been in place for at least 6 years, AFAICT), it's like the same "level" or "type" of change. If we consider it a bug in the README and we just want to inform people, then I a `patch` would work. I thought a `minor` was a stronger signal that some significant change had taken place (also warrants a `minor` per semver.org; not just new features, but significant change in behavior -- that's still backward-compatible), without going to the level of `major`.

I guess it's a fine line. :)

stefcameron:
@thawkin3 Thanks for the updated summary -- so helpful! -- and the additional work. I'll check this out sometime tomorrow.

stefcameron (approved):
This looks great to me, nice work, thank you! 🎉 I'm really happy to have this repo now also using the Testing Library.

stefcameron:
@all-contributors add @thawkin3 for test, infra, doc

stefcameron:
@all-contributors add @idoros for review

thawkin3:
> This looks great to me, nice work, thank you! 🎉 I'm really happy to have this repo now also using the Testing Library.

@stefcameron Thanks! It was fun to work on. I see you've approved this now, so are we good to merge this today? Or are we waiting on the timing for anything else?

stefcameron:
> > This looks great to me, nice work, thank you! 🎉 I'm really happy to have this repo now also using the Testing Library.
> 
> @stefcameron Thanks! It was fun to work on. I see you've approved this now, so are we good to merge this today? Or are we waiting on the timing for anything else?

I'm merging now. I was just pausing just in case there were any last words. 😄

stefcameron:
@all-contributors add @idoros for review

thawkin3:
> I'm merging now. I was just pausing just in case there were any last words. 😄

Haha thanks!

Issue #260, opened 2021-03-04
Add code coverage badge to the README


thawkin3:
Now that the test suite is converted over to use Jest and DOM Testing Library, what do you think about adding a code coverage badge to the README to show off the code coverage?

I see in the `focus-trap-react` repo we used Codecov along with GitHub Actions for the CI. In this `tabbable` repo it's also using GitHub Actions for the CI, so I assume we can generate the code coverage report there and then feed the results to Codecov, right? Or maybe even generate the badge without using Codecov if that's not something you want to do, and we could just use GitHub Actions directly?

Thoughts?

stefcameron:
@thawkin3 Hey, yes, I see no reason why we couldn't add a Codecov badge to this repo now. We should be able to set it up the same way as in `focus-trap-react`. Both repos are setup the same way with GH actions, etc. Just need to invite the Codecov bot to this repo and connect to it with the `codecov` library.

stefcameron:
Codecov bot now has access to this repo. Just need to start feeding it data. 😄

stefcameron:
You're welcome to do the rest if you want. That will just require a small PR. Otherwise, LMK so we don't double-up.

thawkin3:
Thanks! I'm happy to knock this out. I'll have a PR up hopefully in the next day or two to get this set up.

thawkin3:
Here ya go! https://github.com/focus-trap/tabbable/pull/261

PR #261, opened 2021-03-04
ci: adds codecov to CI pipeline and coverage badge to README

thawkin3:
I've added Codecov to the CI Pipeline so that the GitHub Actions workflow will run the unit tests with the coverage report and then pipe the results over to Codecov. I also added a code coverage badge to the README so that we can show off the awesome coverage.

I based my changes off of these two previous MRs in the `focus-trap-react` repo:
- https://github.com/focus-trap/focus-trap-react/pull/227
- https://github.com/focus-trap/focus-trap-react/pull/233

Closes https://github.com/focus-trap/tabbable/issues/260

<details>
<br/>

- [x] Source changes maintain stated browser compatibility.
- ~Unit test coverage added/updated.~
- ~E2E test coverage added/updated.~
- ~Typings added/updated.~
- ~Changeset added (run `yarn changeset` locally to add one, and follow the prompts).~

</details>


package.json
```diff
@@ -65,6 +67,7 @@
     "brfs": "^2.0.2",
     "browserify": "^17.0.0",
     "budo": "^11.6.4",
+    "codecov": "^3.8.1",
```
maraisr, line 70: Rather than include this as a dep, we should _just_ install this in the CI. There zero benefit to ever running this locally. 

So can you please do a `yarn global add codecov` in the gh actions file, and just running that after the tests run. Similarity get rid of line 33 here.

maraisr (changes requested):
Solid effort mate, thank you so much! Just one comment about the coverage commands.

package.json
thawkin3: Love it, thanks! I'll take care of that soon. On that note, we should probably do the same over in the `focus-trap-react` repo since it also has `codecov` listed as a dev dependency: https://github.com/focus-trap/focus-trap-react/blob/master/package.json#L77

I can make the same change over there once we get this MR approved.

yarn.lock
```diff
@@ -331,7 +331,7 @@
     "@babel/helper-plugin-utils" "^7.13.0"
     "@babel/plugin-syntax-logical-assignment-operators" "^7.10.4"
 
-"@babel/plugin-proposal-nullish-coalescing-operator@^7.13.0", "@babel/plugin-proposal-nullish-coalescing-operator@^7.13.8":
+"@babel/plugin-proposal-nullish-coalescing-operator@^7.13.8":
```
thawkin3, line 334: FYI: This was the result after I removed the `codecov` package with `yarn remove codecov`. I checked for what packages are requiring `@babel/plugin-proposal-nullish-coalescing-operator`, and it's only `@babel/preset-env` which does in fact request `@babel/plugin-proposal-nullish-coalescing-operator@^7.13.8`.

So I believe what we see here is correct, and the fact that there were two versions of this package previously installed was unnecessary and that something had gotten out of sync with the lock file.

package.json
thawkin3: After a few attempts, I think I finally have this working correctly. The CI pipeline passes anyway. 😅 

These two issues helped me resolve the problem of the global yarn package not being installed properly or not being installed where it could be found:

- https://github.com/yarnpkg/yarn/issues/648
- https://github.com/actions/setup-node/issues/85

thawkin3:
@maraisr I think this is ready for another round of review whenever you have a minute. Thanks!

stefcameron:
👀

stefcameron (approved):
This looks good to me, thank you @thawkin3 ! I'll let @maraisr chime in before merging.

maraisr (approved):
Thanks for this one! 🦄

thawkin3:
Thanks guys!

PR #286, opened 2021-04-06
display check configuration

idoros:
This PR adds the ability to configure how an element is checked for display. The default behavior is kept as it was, with 2 additional options for an alternative size based check and an opt out option added.

The option `displayCheck` is added to all 4 checking methods of Tabbable.

In order to check the size based option, `getBoundingClientRect` had to be mocked, since Tabbable is checked using jsdom and no longer checked against a real browser. I think this is a miss (I certainly missed this move in the refactor) and we should consider moving back to actual browser testing (against multiple browsers) as this library is trying to integrate with low level browser APIs and mimic browser behavior. 

<details>
<br/>

- Issue that started this PR: #115 .
- Typings added/updated.
- Changeset added

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ Any of the above will _not_ be considered tabbable, though, if any of the follow
 
 - has a negative `tabindex` attribute
 - has a `disabled` attribute
-- either the node itself _or an ancestor of it_ is hidden via `display: none` or `visibility: hidden`
+- either the node itself _or an ancestor of it_ is hidden via `display: none` (*see ["Display check"](#display-check) below to modify this behavior)
+- has `visibility: hidden` style
 - is nested under a closed `<details>` element (with the exception of the first `<summary>` element)
 - is an `<input type="radio">` element and a different radio in its group is `checked`
 
@@ -83,27 +84,49 @@ Type: `boolean`. Default: `false`.
 
 If set to `true`, `rootNode` will be included in the returned tabbable node array, if `rootNode` is tabbable.
 
+##### displayCheck
+
+Type: `full` | `non-zero-area` | `none` . Default: `full`.
+
+configures how to check if an element is displayed, see ["Display check"](#display-check) below.
+
 ### isTabbable
 
 ```js
 import { isTabbable } from 'tabbable';
 
-isTabbable(node);
+isTabbable(node, [options]);
 ```
 
 Returns a boolean indicating whether the provided node is considered tabbable.
+
+#### options
+
+##### displayCheck
+
+Type: `full` | `non-zero-area` | `none` . Default: `full`.
+
+configures how to check if an element is displayed, see ["Display check"](#display-check) below.
 
 ### isFocusable
 
 ```js
 import { isFocusable } from 'tabbable';
 
-isFocusable(node);
+isFocusable(node, [options]);
 ```
 
 Returns a boolean indicating whether the provided node is considered _focusable_.
 
 All tabbable elements are focusable, but not all focusable elements are tabbable. For example, elements with `tabindex="-1"` are focusable but not tabbable.
+
+#### options
+
+##### displayCheck
+
+Type: `full` | `non-zero-area` | `none` . Default: `full`.
+
+configures how to check if an element is displayed, see ["Display check"](#display-check) below.
 
 ### focusable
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -127,16 +150,32 @@ Type: `boolean`. Default: `false`.
...
+
+The `displayCheck` configuration accepts the following options:
+
+- `full`: (default) Most reliably resemble browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (Notice that this doesn't exclude `visibility: hidden` or elements with zero size). This check is by far the slowest option as it might cause layout reflow.
+- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option is much less intensive then the `full` option and better for accessibility as zero-size elements with focusable content are considered a strong accessibility anti-pattern.
+- `none`: This checks completely opt out of the display check. **This option is not recommended**, as it might return elements that are not displayed, and as such not tabbable/focusable and can break accessibility. Make sure you know which elements in your DOM are displayed and can filter them out yourself before using this option.
```
idoros, line 177:
```suggestion
- `none`: This checks completely opt out of the display check. **This option is not recommended**, as it might return elements that are not displayed, and as such not tabbable/focusable and can break accessibility. Make sure you know which elements in your DOM are not displayed and can filter them out yourself before using this option.
```

stefcameron:
@idoros Thank you for doing this! I'll check it out shortly.

stefcameron:
> In order to check the size based option, `getBoundingClientRect` had to be mocked, since Tabbable is checked using jsdom and no longer checked against a real browser. I think this is a miss (I certainly missed this move in the refactor) and we should consider moving back to actual browser testing (against multiple browsers) as this library is trying to integrate with low level browser APIs and mimic browser behavior.

Noted! We can add some Cypress tests to restore actual browser testing, and we could start with a test for this, but doesn't have to be in this PR. focus-trap and focus-trap-react are using Cypress. Easier to maintain 3 repos if they use the same tools. I filed #289 to track this.

README.md
```diff
@@ -127,16 +150,32 @@ Type: `boolean`. Default: `false`.
...
+
+The `displayCheck` configuration accepts the following options:
+
+- `full`: (default) Most reliably resemble browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (Notice that this doesn't exclude `visibility: hidden` or elements with zero size). This check is by far the slowest option as it might cause layout reflow.
+- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option is much less intensive then the `full` option and better for accessibility as zero-size elements with focusable content are considered a strong accessibility anti-pattern.
+- `none`: This checks completely opt out of the display check. **This option is not recommended**, as it might return elements that are not displayed, and as such not tabbable/focusable and can break accessibility. Make sure you know which elements in your DOM are not displayed and can filter them out yourself before using this option.
```
stefcameron, line 177:
```suggestion
- `none`: This completely opts out of the display check. **This option is not recommended**, as it might return elements that are not displayed, and as such not tabbable/focusable and can break accessibility. Make sure you know which elements in your DOM are not displayed and can filter them out yourself before using this option.
```

Just a typo.

test/focusable.test.js
```diff
@@ -293,6 +294,68 @@ describe('focusable', () => {
           expectedFocusableIds
         );
       });
+
+      describe('displayed check', () => {
```
stefcameron, line 298: Maybe this was copy/paste, but could you find/replace "Tabbable" with "Focusable" in this whole block? Just confusing to reach variable names with "Tabbable", yet we're testing `focusable()` here. 😄

test/isFocusable.test.js
```diff
@@ -298,4 +298,96 @@ describe('isFocusable', () => {
...
+      mockElementsSizes(container);
+      document.body.append(container);
+      const displayedTop = getByTestId(container, 'displayed-top');
+      const displayedNested = getByTestId(container, 'displayed-nested');
+      const displayedZeroSize = getByTestId(container, 'displayed-zero-size');
+      const displayedNoneTop = getByTestId(container, 'displayed-none-top');
+      const nestedUnderDisplayedNone = getByTestId(
+        container,
+        'nested-under-displayed-none'
+      );
```
stefcameron, line 334: This block is repeated in each test. Could we move this to a `beforeEach()` to simplify the code?

test/isTabbable.test.js
```diff
@@ -293,4 +293,96 @@ describe('isTabbable', () => {
...
+      mockElementsSizes(container);
+      document.body.append(container);
+      const displayedTop = getByTestId(container, 'displayed-top');
+      const displayedNested = getByTestId(container, 'displayed-nested');
+      const displayedZeroSize = getByTestId(container, 'displayed-zero-size');
+      const displayedNoneTop = getByTestId(container, 'displayed-none-top');
+      const nestedUnderDisplayedNone = getByTestId(
+        container,
+        'nested-under-displayed-none'
+      );
```
stefcameron, line 329: Same here: Can we move to a `beforeEach()` block to simplify the test code in here?

stefcameron (changes requested):
Looking good, thank you!! A few minor requests. 😄

test/focusable.test.js
idoros: good catch!

test/isFocusable.test.js
idoros: unify under a setup function (beforeEach doesn't have a context that it returns to the test and I didn't want to keep so many variables in the describe closure).

test/isTabbable.test.js
idoros: unify under a setup function

stefcameron (approved):
Looks good! Thanks for the additional fixes. I'll get this merged and published.

Fixes #115 

I also logged https://github.com/focus-trap/focus-trap/issues/362 to track the work of surfacing this new issue over on focus-trap.

Issue #50, opened 2020-03-25
Tab Across Web Components


scottfr:
It would be great if tabbable could identify the tab ordering between input fields across multiple web components.

Unfortunately since query selectors can't span shadow roots, it may require a very different approach.

stefcameron:
> For me personally a solution can work, where I iterate over the DOM nodes and an API could tell me, whether that Node is tabbable or not.

This library does now have `isTabbable()` and `isFocusable()` APIs that might help.

idoros:
Unfortunately current solution is not enough as Tabbable code doesn't deal with the concept of shadow dom at all. 

This is what needs to be solved as far as I can see: 

## Query shadow roots 

- locate elements with shadow roots
- query internal shadow doms recursively and insert potential nested elements in the correct order into the candidate list

The problem is that as far as I know, there is no CSS selector to query elements that has a shadow root. That means that a search needs to iterate on every element to check for `.shadowRoot` property and then query it internally. We could provide an attribute to mark elements with shadow root (e.g. `data-tabbable-root`), and a utility function to search and mark such elements, allowing a user to mark them themselves or run the function as needed.

## Filter elements within shadow dom

- `parentElement` of top level elements in a shadow root is null - need this to figure out if an element is displayed (maybe `parentNode.host`)
- when searching for other radio elements with no containing form, the fallback should be to the shadow root and not the document (if one exist)

-----

No matter the solution, currently Tabbable is tested with jsdom and implementing this without testing in a real browser is problematic.

LohitakshTrehan:
The approach we followed was BFS traversal of the tree if there is a shadow root element. We used jQuery `$(:tabbable)` selector on each element to find if it is tabbable or not and pushed it in array if it is a tabbable element. After that we compared the index of current focused element to the next tabbable element in array. and transfer the focus there. If it was the last, then we started on first array element. 
It used a lot of computation. But we needed that feature in our project. There was no visible delay for user.

idoros:
> BFS traversal of the tree if there is a shadow root element

That is what I was suggesting, but I wouldn't run this by default for every tabbable/focusable call as not all users use shadow-dom (at list not in the current major version, maybe with a flag that can be flipped in the future).

> There was no visible delay

That depends on the size and complexity of the DOM you have, the device your code is running on, and how frequent you run your checks.

LohitakshTrehan:
Yes we can give a flag.
[...]

PR #304, opened 2021-04-23
shadow dom support

idoros:
This PR adds support for shadow dom (as requested in #50). Currently I just placed all tests in a single e2e spec, because it was much easier to handle, I will probably break it down into the 4 apis before we merge. The code is separated into 2 parts:

1. changes to check isTabbable/isFocusable
2. alternative to dom query which walks the dom 

new options API:
- getShadowRoot(element): called for any element without `shadowRoot` property in order to detect shadow dom. (return `shadowRoot` in order to query internally, `true` to indicate there is a shadow-root, but it's not available or `false` for no shadow-root)

- [x] dev support for fixtures with [declarative shadow dom](https://web.dev/declarative-shadow-dom/):
    - add development shadow-root-utils to handle hydration, polyfill and expose closed shadow root for tests.
    - add options to render a subtree of a fixture `setupFixture(content, {**caseId**})`.
    - change debug page to include and work with shadow dom examples.
- support isTabbable/isFocusable:
    - [x] separate radio buttons group between each light/shadow (found [unexpected behavior](https://bugs.chromium.org/p/chromium/issues/detail?id=1200261) in chromium).
    - [x] check for display in shadow dom and across shadow boundary when possible:
        - if a shadow root is detected, but cannot be accessed, then fallback to zero-size display check.
        - **caveat**: if an element, or one of it's ancestors, is nested under an undetected shadow-root with `display=none` or is not slotted, then the element will be wrongly considered focusable (only for `displayCheck: "full"`).
- [x] support scanning dom via walking instead of query
    - when `getShadowRoot` option is supplied then walking is picked over query and shadow-roots and slots are visited
    - new type of candidate list with scoped items is supported. a scope item is not a tabbable/focusable item by itself, but contains a list of nested elements that needs to be sorted between themselves and inserted according to the scoped item computed position (shadow-roots and slots generates scoped items, while shadow hosts can also be focusable/tabbable themselves).
    - sort now accepts the new scoped candidate list, sort and flatten it.
- `delegatesFocus` is not implemented here.
    - ~I played with it, but it only works in chromium (can't get it to [work on Firefox](https://bug-166484-attachments.webkit.org/attachment.cgi?id=297772), although MDN states [differently](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus#browser_compatibility)), and in any case it's marked as [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus)~ [not implemented](https://bugzilla.mozilla.org/show_bug.cgi?id=1706275) in Firefox
    - if this were to be added then it's simply a matter of adding a check in `isTabbable/isFocusable`. The check try to find `shadowRoot.delegatesFocus` (currently exposed on chromium), or ask for `options.delegatesFocus(node)` and allow whoever uses Tabbable to provide the flag according to available shadow knowledge and  browser support.
    - added support for `shadowrootdelegatesfocus` in the declarative shadow dom dev utils for now
- [x] types
- [ ] documentation -  README updated (API changes, instructions, etc.).
- [ ] make sure source is compatible with the acceptable ECMAScript version
- [ ] Changeset
    
Some thoughts:
- closed mode shadow roots cannot be accessed without help from whoever defined them. This can be solved with a component framework/platform or apparently anyone with access to the page before the components are defined by overriding `customElements.define` or `Element.prototype.attachShadow` 🤔
- it might be helpful to offer a way to mark dom elements and run a query instead of walking, but that's just optimizations features that can be added later.
- now that there is an iterative dom walk option, iframes content can be scanned in a similar way.
- change debug page to render one fixture/fixture case at a time, currently tabindex between fixtures make it hard to use.
- maybe change source to latest ECMAScript to clean it up and just transpile for legacy versions.
- also it would be nice to break down the source into separate files as its hard to navigate through.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -244,7 +244,7 @@ const isZeroArea = function (node) {
   const { width, height } = node.getBoundingClientRect();
   return width === 0 && height === 0;
 };
-const isHidden = function (node, options) {
+const isHidden = function (node, { displayCheck, getShadowRoot = noop }) {
   if (getComputedStyle(node).visibility === 'hidden') {
     return true;
   }
@@ -254,8 +254,7 @@ const isHidden = function (node, options) {
   if (matches.call(nodeUnderDetails, 'details:not([open]) *')) {
     return true;
   }
-  const displayCheck = options.displayCheck;
-  const getShadowRoot = options.getShadowRoot || noop;
+
   if (!displayCheck || displayCheck === 'full') {
     while (node) {
       if (getComputedStyle(node).display === 'none') {
--- a/src/index.js
+++ b/src/index.js
@@ -100,7 +100,8 @@ const isTabbableRadio = function (node) {
   if (!node.name) {
     return true;
   }
-  const radioScope = node.form || node.ownerDocument;
+  const radioScope =
+    node.form || node.getRootNode ? node.getRootNode() : node.ownerDocument;
 
   const queryRadios = function (name) {
     return radioScope.querySelectorAll(
--- a/src/index.js
+++ b/src/index.js
@@ -4,7 +4,7 @@ const candidateSelectors = [
   'textarea',
   'a[href]',
   'button',
-  '[tabindex]',
+  '[tabindex]:not(slot)',
   'audio[controls]',
   'video[controls]',
   '[contenteditable]:not([contenteditable="false"])',
--- a/src/index.js
+++ b/src/index.js
@@ -37,11 +37,69 @@ const getCandidates = function (el, includeContainer, filter) {
   return candidates;
 };
 
+const getCandidatesIteratively = function (
+  elements,
+  includeContainer,
+  options
+) {
+  const candidates = [];
+  const elementsToCheck = Array.from(elements);
+  while (elementsToCheck.length) {
+    const element = elementsToCheck.shift();
+    if (element.tagName === 'SLOT') {
+      // add shadow dom slot scope (slot itself cannot be focusable)
+      const assigned = element.assignedElements();
+      const content = assigned.length ? assigned : element.children;
+      const nestedCandidates = getCandidatesIteratively(content, true, options);
+      if (options.flatten) {
+        candidates.push(...nestedCandidates);
+      } else {
+        candidates.push({
+          scope: element,
+          candidates: nestedCandidates,
+        });
+      }
+    } else {
+      // check candidate element
+      const validCandidate = matches.call(element, candidateSelector);
+      if (
+        validCandidate &&
+        options.filter(element) &&
+        (includeContainer || !elements.includes(element))
+      ) {
+        candidates.push(element);
+      }
+      // iterate over content
+      const shadowRoot = element.shadowRoot || options.getShadowRoot(element);
+      if (shadowRoot) {
+        // add shadow dom scope
+        const nestedCandidates = getCandidatesIteratively(
+          shadowRoot === true ? element.children : shadowRoot.children,
+          true,
+          options
+        );
+        if (options.flatten) {
+          candidates.push(...nestedCandidates);
+        } else {
+          candidates.push({
+            scope: element,
+            candidates: nestedCandidates,
+          });
+        }
+      } else {
+        // add light dom scope
+        elementsToCheck.unshift(...element.children);
+      }
+    }
+  }
+  return candidates;
+};
+
 const isContentEditable = function (node) {
   return node.contentEditable === 'true';
 };
 
-const getTabindex = function (node) {
+const getTabindex = function (node, isScope) {
   const tabindexAttr = parseInt(node.getAttribute('tabindex'), 10);
 
   if (!isNaN(tabindexAttr)) {
@@ -59,8 +117,13 @@ const getTabindex = function (node) {
   //  yet they are still part of the regular tab order; in FF, they get a default
   //  `tabIndex` of 0; since Chrome still puts those elements in the regular tab
   //  order, consider their tab index to be 0.
+  //
+  // isScope is positive for custom element with shadow root or slot that by default
+  // have tabIndex -1, but need to be sorted by document order in order for their
+  // content to be inserted in the correct position
   if (
-    (node.nodeName === 'AUDIO' ||
+    (isScope ||
+      node.nodeName === 'AUDIO' ||
       node.nodeName === 'VIDEO' ||
       node.nodeName === 'DETAILS') &&
     node.getAttribute('tabindex') === null
@@ -263,47 +326,77 @@ const isNodeMatchingSelectorTabbable = function (options, node) {
   return true;
 };
 
-const tabbable = function (el, options) {
-  options = options || {};
-
+const sortByOrder = function (candidates) {
   const regularTabbables = [];
   const orderedTabbables = [];
-
-  const candidates = getCandidates(
-    el,
-    options.includeContainer,
-    isNodeMatchingSelectorTabbable.bind(null, options)
-  );
-
-  candidates.forEach(function (candidate, i) {
-    const candidateTabindex = getTabindex(candidate);
+  candidates.forEach(function (item, i) {
+    const isScope = !!item.scope;
+    const element = isScope ? item.scope : item;
+    const candidateTabindex = getTabindex(element, isScope);
+    const elements = isScope ? sortByOrder(item.candidates) : element;
     if (candidateTabindex === 0) {
-      regularTabbables.push(candidate);
+      isScope
+        ? regularTabbables.push(...elements)
+        : regularTabbables.push(element);
     } else {
       orderedTabbables.push({
         documentOrder: i,
         tabIndex: candidateTabindex,
-        node: candidate,
+        item: item,
+        isScope: isScope,
+        content: elements,
       });
     }
   });
 
-  const tabbableNodes = orderedTabbables
+  return orderedTabbables
     .sort(sortOrderedTabbables)
-    .map((a) => a.node)
+    .reduce((acc, sortable) => {
+      sortable.isScope
+        ? acc.push(...sortable.content)
+        : acc.push(sortable.content);
+      return acc;
+    }, [])
     .concat(regularTabbables);
+};
 
-  return tabbableNodes;
+const tabbable = function (el, options) {
+  options = options || {};
+
+  let candidates;
+  if (options.getShadowRoot) {
+    candidates = getCandidatesIteratively([el], options.includeContainer, {
+      filter: isNodeMatchingSelectorTabbable.bind(null, options),
+      flatten: false,
+      getShadowRoot: options.getShadowRoot,
+    });
+  } else {
+    candidates = getCandidates(
+      el,
+      options.includeContainer,
+      isNodeMatchingSelectorTabbable.bind(null, options)
+    );
+  }
+  return sortByOrder(candidates);
 };
 
 const focusable = function (el, options) {
   options = options || {};
 
-  const candidates = getCandidates(
-    el,
-    options.includeContainer,
-    isNodeMatchingSelectorFocusable.bind(null, options)
-  );
+  let candidates;
+  if (options.getShadowRoot) {
+    candidates = getCandidatesIteratively([el], options.includeContainer, {
+      filter: isNodeMatchingSelectorFocusable.bind(null, options),
+      flatten: true,
+      getShadowRoot: options.getShadowRoot,
+    });
+  } else {
+    candidates = getCandidates(
+      el,
+      options.includeContainer,
+      isNodeMatchingSelectorFocusable.bind(null, options)
+    );
+  }
 
   return candidates;
 };
--- a/src/index.js
+++ b/src/index.js
@@ -13,12 +13,18 @@ const candidateSelectors = [
 ];
 const candidateSelector = /* #__PURE__ */ candidateSelectors.join(',');
 
-const matches =
-  typeof Element === 'undefined'
-    ? function () {}
-    : Element.prototype.matches ||
-      Element.prototype.msMatchesSelector ||
-      Element.prototype.webkitMatchesSelector;
+const NoElement = typeof Element === 'undefined';
+
+const matches = NoElement
+  ? function () {}
+  : Element.prototype.matches ||
+    Element.prototype.msMatchesSelector ||
+    Element.prototype.webkitMatchesSelector;
+
+const getRootNode =
+  !NoElement && Element.prototype.getRootNode
+    ? (element) => element.getRootNode()
+    : (element) => element.ownerDocument;
 
 const getCandidates = function (el, includeContainer, filter) {
   let candidates = Array.prototype.slice.apply(
@@ -100,9 +106,7 @@ const isTabbableRadio = function (node) {
   if (!node.name) {
     return true;
   }
-  const radioScope =
-    node.form || node.getRootNode ? node.getRootNode() : node.ownerDocument;
-
+  const radioScope = node.form || getRootNode(node);
   const queryRadios = function (name) {
     return radioScope.querySelectorAll(
       'input[type="radio"][name="' + name + '"]'
@@ -141,7 +145,12 @@ const isNonTabbableRadio = function (node) {
   return isRadio(node) && !isTabbableRadio(node);
 };
 
-const isHidden = function (node, displayCheck) {
+const noop = () => {};
+const isZeroArea = function (node) {
+  const { width, height } = node.getBoundingClientRect();
+  return width === 0 && height === 0;
+};
+const isHidden = function (node, options) {
   if (getComputedStyle(node).visibility === 'hidden') {
     return true;
   }
@@ -151,16 +160,35 @@ const isHidden = function (node, displayCheck) {
   if (matches.call(nodeUnderDetails, 'details:not([open]) *')) {
     return true;
   }
+  const displayCheck = options.displayCheck;
+  const getShadowRoot = options.getShadowRoot || noop;
   if (!displayCheck || displayCheck === 'full') {
     while (node) {
       if (getComputedStyle(node).display === 'none') {
         return true;
       }
-      node = node.parentElement;
+      const parentElement = node.parentElement;
+      const rootNode = getRootNode(node);
+      if (
+        parentElement &&
+        !parentElement.shadowRoot &&
+        getShadowRoot(parentElement)
+      ) {
+        // fallback to zero area size for unreachable shadow dom
+        return isZeroArea(node);
+      } else if (node.assignedSlot) {
+        // iterate up slot
+        node = node.assignedSlot;
+      } else if (!parentElement && rootNode !== node.ownerDocument) {
+        // cross shadow boundary
+        node = rootNode.host;
+      } else {
+        // iterate up normal dom
+        node = parentElement;
+      }
     }
   } else if (displayCheck === 'non-zero-area') {
-    const { width, height } = node.getBoundingClientRect();
-    return width === 0 && height === 0;
+    return isZeroArea(node);
   }
 
   return false;
@@ -214,7 +242,7 @@ const isNodeMatchingSelectorFocusable = function (options, node) {
   if (
     node.disabled ||
     isHiddenInput(node) ||
-    isHidden(node, options.displayCheck) ||
+    isHidden(node, options) ||
     // For a details element with a summary, the summary element gets the focus
     isDetailsWithSummary(node) ||
     isDisabledFromFieldset(node)
--- a/src/index.js
+++ b/src/index.js
@@ -26,6 +26,12 @@ const getRootNode =
     ? (element) => element.getRootNode()
     : (element) => element.ownerDocument;
 
+/**
+ * @param {Element} el container to check in
+ * @param {boolean} includeContainer add container to check
+ * @param {(node: Element) => boolean} filter filter candidates
+ * @returns {Element[]}
+ */
 const getCandidates = function (el, includeContainer, filter) {
   let candidates = Array.prototype.slice.apply(
     el.querySelectorAll(candidateSelector)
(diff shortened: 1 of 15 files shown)


idoros:
@stefcameron, I added this PR in order to see what can be done for shadow doms, I'm open for suggestions and willing to break it down or change if you prefer.

@scomea, it would be great if you can check if this helps your issue

stefcameron:
> @stefcameron, I added this PR in order to see what can be done for shadow doms, I'm open for suggestions and willing to break it down or change if you prefer.

Thanks for getting this started! I will take a look as soon as I can.

stefcameron:
I haven't forgotten about this... Slowly trying to grok it all.

src/index.js
```diff
@@ -31,11 +37,69 @@ const getCandidates = function (el, includeContainer, filter) {
   return candidates;
 };
 
+const getCandidatesIteratively = function (
+  elements,
+  includeContainer,
+  options
+) {
```
stefcameron, line 44: For future maintenance, it would be really helpful to have a JSDoc comment on this function that details what the parameters are -- especially `options` and all possible properties. It's a bit too hard to decipher it easily just from reading the code.

src/index.js
```diff
@@ -31,11 +37,69 @@ const getCandidates = function (el, includeContainer, filter) {
...
+      const validCandidate = matches.call(element, candidateSelector);
+      if (
+        validCandidate &&
+        options.filter(element) &&
+        (includeContainer || !elements.includes(element))
+      ) {
+        candidates.push(element);
+      }
+      // iterate over content
+      const shadowRoot = element.shadowRoot || options.getShadowRoot(element);
```
stefcameron, line 73: We should probably default `options` to have a noop function for `getShadowRoot` option if it isn't defined. Otherwise, if you have an element with a closed shadow, this code will crash, right?

src/index.js
```diff
@@ -31,11 +37,69 @@ const getCandidates = function (el, includeContainer, filter) {
...
+          scope: element,
+          candidates: nestedCandidates,
+        });
+      }
+    } else {
+      // check candidate element
+      const validCandidate = matches.call(element, candidateSelector);
+      if (
+        validCandidate &&
+        options.filter(element) &&
```
stefcameron, line 67: Better default `options.filter = (element) => element` to avoid crash.

src/index.js
stefcameron: That should also include details about `getShadowRoot()` option and its possible return values, since it looks like it can return probably either a boolean or an Element?

src/index.js
```diff
@@ -150,16 +223,35 @@ const isHidden = function (node, displayCheck) {
   if (matches.call(nodeUnderDetails, 'details:not([open]) *')) {
     return true;
   }
+  const displayCheck = options.displayCheck;
+  const getShadowRoot = options.getShadowRoot || noop;
```
stefcameron, line 227: Could be written with more modern destructuring like:
```javascript
const {
  displayCheck,
  getShadowRoot = noop
} = options;
```

stefcameron (changes requested):
Wow, there's some bit of work in here already! For now, I focused on the main source changes, but I don't think I'm the best to provide a thorough review since there are a few concepts here that are still quite new to me.

One of my main concerns is how these changes will affect `focus-trap`. We need to make sure changes here will not break it since it fully depends on `tabbable`'s output, so please incorporate integration testing with `focus-trap` as you're making these changes (not saying to add tests for `focus-trap` here, but to locally test `focus-trap` with `yarn link` to make sure it's still running smoothly).

I'm wondering if @Andarist has any experience with Shadow DOM (something tells me he probably does) and might have some cycles to provide a review here as well.

stefcameron:
AFAICT so far, the changes seem to be covering all the good cases we discussed in #50, including open/closed shadows and slotted elements, and now also declarative templates.

We should probably add some tests to `focus-trap` that stress the shadow DOM to make sure the integration works.

> maybe change source to latest ECMAScript to clean it up and just transpile for legacy versions.

The code should be getting transpiled correctly, so we should be able to continue modernizing the source.

> also it would be nice to break down the source into separate files as its hard to navigate through.

Sure, though maybe not in this PR as it could make it harder than it already is to review (at least for me).

src/index.js
idoros: will add jsdocs.

No sure `getShadowRoot` full description belongs here, as it is used in other flows as well, and is part of the external api, I will find a place for it that will hopefully be clear.

src/index.js
idoros: filter is not optional, it's always provided internally by `tabbable/focusable` in the same way it is provided to the original `getCandidates`

src/index.js
idoros: definitely 👍

src/index.js
idoros: 👍

idoros:
>  how these changes will affect focus-trap. We need to make sure changes here will not break it

The main query flow is unchanged, so I don't see how it can break anything. 
there are 2 other changes in filtering:
1. isTabbable: radio button searches in the shadow root instead of the document if available - I would consider this a fix
2. isFocusable: display check that tries its best to check for display under nested shadow roots - also a fix.

> We should probably add some tests to focus-trap that stress the shadow DOM to make sure the integration works...We should probably add some tests to focus-trap that stress the shadow DOM to make sure the integration works.

I will test locally like you suggested, if you want to release a beta version of tabbable then I don't mind opening a PR there to test it out. I think ideally you would run all tests under the 2 different query flows once you expose the option to provide `getShadowRoot`. 

> I'm wondering if @Andarist has any experience with Shadow DOM

Definitely, please review as much as possible as I'm not an expert on shadow roots or web components (except for following the  standards, I never actually built anything substantial with them and have no big project to test this on).

I will keep modernization and splits to another PR in anycase. I was just pointing out things as I wrote the code...

src/index.js
stefcameron: ok

src/index.js
stefcameron: I see. `options` is fine. I think it'd still have the same questions were it named `config`. JSDoc will clarify as it'll show non-optional.

stefcameron:
> I will test locally like you suggested, if you want to release a beta version of tabbable then I don't mind opening a PR there to test it out. I think ideally you would run all tests under the 2 different query flows once you expose the option to provide `getShadowRoot`.

Releasing a beta is also something I could do. Let's see if we get more reviews first, then we can decide on the best course of action here (local testing with links or beta releases).

idoros:
I tested `focus-trap` locally against the changes and it passes. I also tested both projects with a modified default value for `getShadowRoot` (returns false) in order to check all the tests against the alternative query flow and both passed.

stefcameron:
> I tested `focus-trap` locally against the changes and it passes. I also tested both projects with a modified default value for `getShadowRoot` (returns false) in order to check all the tests against the alternative query flow and both passed.

@idoros Thanks for those JSDoc updates, checking over on `focus-trap`, and checking the new tests here.

I'm not sure if @Andarist will find time to review this or how long we should wait.

Maybe it's time I publish this as a beta (and `focus-trap` beta that uses this beta, I guess) and have @LohitakshTrehan give it a try on some of his code? Maybe some people working on https://github.com/microsoft/fast/issues/4035 could also try it?

WDYT?

scomea:
> Maybe it's time I publish this as a beta (and `focus-trap` beta that uses this beta, I guess) and have @LohitakshTrehan give it a try on some of his code? Maybe some people working on [microsoft/fast#4035](https://github.com/microsoft/fast/issues/4035) could also try it?

I've been meaning to give this a try.  On my short list.

stefcameron:
OK, LMK if a beta is needed, or if you'll just `yarn/npm link` it locally.

Andarist:
> I'm not sure if @Andarist will find time to review this or how long we should wait.

Unfortunately, I don't have much experience with Shadow DOM... and looking at the added complexity here I don't regret it 😉 The fact that this also just can't handle all use cases with the Shadow DOM (because of its design) ain't helping me to switch over to web components.

stefcameron:
> > I'm not sure if @Andarist will find time to review this or how long we should wait.
> 
> Unfortunately, I don't have much experience with Shadow DOM... and looking at the added complexity here I don't regret it 😉 The fact that this also just can't handle all use cases with the Shadow DOM (because of its design) ain't helping me to switch over to web components.

Thanks, @Andarist, I appreciate your thoughts on it!

scomea:
Wasn't able to get the local link to work directly in the Fast repo (building/packaging isn't my strong suit) but I do have the branch itself running.  Seems to work at first glance.   Gonna dig in some more and try some things, so actively looking at it.

stefcameron:
> Wasn't able to get the local link to work directly in the Fast repo (building/packaging isn't my strong suit) but I do have the branch itself running. Seems to work at first glance. Gonna dig in some more and try some things, so actively looking at it.

@scomea Wonderful, we appreciate your help on validation!

scomea:
@idoros Thanks for digging into this very difficult problem!

I have a local version running that I can link to in my local fast repo.  I hacked some extra buttons into the shadow dom of our (not "closed") dialog and used this call to tabbable:

this.tabbableElements = tabbable(this, { getShadowRoot: () => true });

I'm fuzzy on how getShadowRoot works so I'm kinda winging it on that with my goofy callback.  

Here's what my Dom looks like:

image

The 'tabbable' call above returns the three buttons hidden in the shadow dom (yay!) but does not find the elements in the slot. Those do get found without the 'getShadow' option.  I guess there may be some slot scanning that should be happening?

Another wrinkle is that even if tabbable returns elements in the shadow dom there is another issue in that api's like document.activeElement and event.target report the shadow parent rather than the element in it's shadow dom that is actually focused (boo).  Not sure how to approach that or the "closed" shadow dom issue.  

So far I've been trying to keep all the focusable bits of any complex (ie. not single focus) web component I've built in slots so they are in the light dom.  I think I'm going to keep doing that.

idoros:
> I'm fuzzy on how getShadowRoot works so I'm kinda winging it on that with my goofy callback.

`getShadowRoot` needs to get the actual shadowRoot in order to query it's internals for focusable/tabbable elements. returning true hints the other flow (`isTabbable/isFocusable`). if you don't know if an element has a shadowRoot then don't automatically return true.

If all your attached shadows are open then it should work find without any input from `getShadowRoot`.

> The 'tabbable' call above returns the three buttons hidden in the shadow dom (yay!) but does not find the elements in the slot. Those do get found without the 'getShadow' option. I guess there may be some slot scanning that should be happening?

The native buttons in the shadow dom are found as the gallery is open. The slotted elements are in the light dom and should be found in either condition. In your case the custom buttons have `tabindex` and should be returned, and in case they also contain open shadow roots then any additional tabbable element inside them should also be included.

some questions about your setup:
- do the `fast-button/checkbox` have an open shadow root?
 - can you try placing a native button/input in a slot and see they are returned?
 - do you have some branch that I can see and test this scenario?

> api's like document.activeElement and event.target report the shadow parent rather than the element in it's shadow dom

That is by design as far as I know and not related to Tabbable, `document.activeElement` points to the element in the light dom that is or has inner focus, in case there is a shadow element focused inside the host element, then the shadowRoot has it's own `shadowRoot.activeElement` (you can [see how I loop over the activeElements in the debug page here](https://github.com/focus-trap/tabbable/pull/304/files#diff-9f59f8ded8585de988c1dfaa517e067f8d9a76696125d695249c4ed82cecddd7R23)). With `eventTarget` you also get the element in the scope that you listen to. 

> So far I've been trying to keep all the focusable bits of any complex (ie. not single focus) web component I've built in slots so they are in the light dom. I think I'm going to keep doing that.

Not sure that this scales. I think in most cases, keeping the shadowRoot open is the best approach.

scomea:
> `getShadowRoot` needs to get the actual shadowRoot in order to query it's internals for focusable/tabbable elements. returning true hints the other flow (`isTabbable/isFocusable`). if you don't know if an element has a shadowRoot then don't automatically return true.
> 
> If all your attached shadows are open then it should work find without any input from `getShadowRoot`.

So getShadowRoot is intended for devs who are able to identify components with closed shadows and have a way to return the actual root?  

> * do the `fast-button/checkbox` have an open shadow root?

Yes, but I'm thinking they should be 'closed' (I believe the fluent versions are) because your scan returns both the top level button component as well as the actual button hidden in its shadow.  They should be one thing.

> * can you try placing a native button/input in a slot and see they are returned?

I added a fast button to the shadow button group and a native button to the slotted group.  When I call tabbable like so:
 ```const tabbableShadow: (HTMLElement | SVGElement)[] = tabbable(this, { getShadowRoot: () => true });```
I get all the elements in the shadow button group.

When I call it like so:
``` const tabbableReg: (HTMLElement | SVGElement)[] = tabbable(this);```
I get all the elements in the slotted button group.

Additionally.  If I set my dialog component's shadow mode to closed both queries return the slotted group. 
(Let's ignore that fast-buttons return as two, I know what's going on there.)

> * do you have some branch that I can see and test this scenario?

Here is my hack branch of the fast repo, I yarn linked it to use a local instance of your branch:
https://github.com/microsoft/fast/tree/users/scomea/tabbable-shadow-testing

You want to `yarn install` at the repo root, then `yarn start` in 'fast components' 
https://github.com/microsoft/fast/tree/users/scomea/tabbable-shadow-testing/packages/web-components/fast-components

The base code for dialog where I've been hacking at are here.  (I've been putting breakpoints in dialog.ts to see what tabbable finds):
https://github.com/microsoft/fast/blob/4f7a5ac399c698763c9b2091cbdc71c8dc21ce2f/packages/web-components/fast-foundation/src/dialog/dialog.ts#L191
https://github.com/microsoft/fast/blob/users/scomea/tabbable-shadow-testing/packages/web-components/fast-foundation/src/dialog/dialog.template.ts

And the component is declared here:
https://github.com/microsoft/fast/blob/users/scomea/tabbable-shadow-testing/packages/web-components/fast-components/src/dialog/index.ts

> > api's like document.activeElement and event.target report the shadow parent rather than the element in it's shadow dom
> 
> That is by design as far as I know and not related to Tabbable,

Yes, not related to tabbable, but typically part of the focus trap scenario we're ultimately trying to solve.

> > So far I've been trying to keep all the focusable bits of any complex (ie. not single focus) web component I've built in slots so they are in the light dom. I think I'm going to keep doing that.
> 
> Not sure that this scales. I think in most cases, keeping the shadowRoot open is the best approach.

It does make things more complicated than simply plopping down some buttons in a template and calling your component good, that's for sure.  But so far it has worked for me.

**Comment by idoros** [comment]
> So getShadowRoot is intended for devs who are able to identify components with closed shadows and have a way to return the actual root?

Yes, in case you are in a position of a framework (or have access to the page before load) that has access to the root of closed shadow components you can provide it so that Tabbable can walk the inner dom and return the correct order (notice that if you expose the result outside the scope of your code then you potentially expose the closed dom). I thought that "fast" is in good position to keep track of the inner roots of close components and use it for a better internal focus trap.

I run your setup and as all of your components are open (dialog/button/checkbox), then `getShadowRoot` is only used as flag to collect dom across shadow boundaries, and simply returning `true` to everything makes every element into a "scoped item" which means its content is not sorted with the outside of its scope.

In your case that everything is open, calling `tabbable(dialog, {getShadowRoot: (node) => {/*return nothing as nothing is known*/}})` returns the following:
image
which seems to contain everything in the correct order.

> I'm thinking they should be 'closed' (I believe the fluent versions are) because your scan returns both the top level button component as well as the actual button hidden in its shadow. They should be one thing

That is a decision for "fast", tabbable can just try it's best to return accurate results as they are found. you can close your component, add `tabindex`, and return `true` in `getShadowRoot` for best results in that case (if `delegatesFocus` ever becomes a thing then we can auto pick it up or provide a mechanism to provide the info for an element and you wouldn't need the `tabindex`).

**Comment by scomea** [comment]
> > So getShadowRoot is intended for devs who are able to identify components with closed shadows and have a way to return the actual root?
> 
> Yes, in case you are in a position of a framework (or have access to the page before load) that has access to the root of closed shadow components you can provide it so that Tabbable can walk the inner dom and return the correct order (notice that if you expose the result outside the scope of your code then you potentially expose the closed dom). I thought that "fast" is in good position to keep track of the inner roots of closed components and use it for a better internal focus trap.
> 
> I run your setup and as all of your components are open (dialog/button/checkbox), then `getShadowRoot` is only used as flag to collect dom across shadow boundaries, and simply returning `true` to everything makes every element into a "scoped item" which means its content is not sorted with the outside of its scope.
> 
> In your case that everything is open, calling `tabbable(dialog, {getShadowRoot: (node) => {/*return nothing as nothing is known*/}})` returns the following:
> image
> which seems to contain everything in the correct order.

I'm not sure how useful the getShowRoot() approach is in production scenarios, how many developers can actually provide the shadow root if they don't have some kind of special access to the component?

**Comment by scomea** [comment]
> > I'm thinking they should be 'closed' (I believe the fluent versions are) because your scan returns both the top level button component as well as the actual button hidden in its shadow. They should be one thing
> 
> That is a decision for "fast", tabbable can just try it's best to return accurate results as they are found. you can close your component, add `tabindex`, and return `true` in `getShadowRoot` for best results in that case (if `delegatesFocus` ever becomes a thing then we can auto pick it up or provide a mechanism to provide the info for an element and you wouldn't need the `tabindex`).

I say that because we have some components like 'fast-button' that are focusable and wrap a single focusable native element in their shadow dom.  That needs to resolve as a single focus target.

**Comment by Paddy613** [comment]
>

**Comment by justinfagnani** [comment]
@idoros I don't know where you're seeing that `delegatesFocus` is deprecated, but I'm pretty sure it's not.

**Comment by idoros** [comment]
Used to be stated as deprecated on mdn when this was written:



https://github.com/mdn/content/commits/main/files/en-us/web/api/shadowroot/delegatesfocus/index.html

It's been a while, but I'm happy to rethink/implement this into the api if someone would show any interest in accepting it.  (I will add an update to the original comment tomorrow)

**Comment by stefcameron** [comment]
> It's been a while, but I'm happy to rethink/implement this into the api if someone would show any interest in accepting it. I updated the notice about deprecation in the original comment.

I'm waiting for others more knowledgeable in the space to review. I believe we were in the middle of a discussion with @scomea AFAICT. I'm willing to accept the PR if things look solid. There are a couple of minor items that still need to be done (unchecked items in the PR description), but those are best left until the tail-end of the review.

**Comment by scomea** [comment]
> I'm waiting for others more knowledgeable in the space to review. I believe we were in the middle of a discussion with @scomea AFAICT. I'm willing to accept the PR if things look solid. There are a couple of minor items that still need to be done (unchecked items in the PR description), but those are best left until the tail-end of the review.

My reservation is mostly about the usefulness of the closed shadow dom mechanism as it seems to require special access to those components that most devs won't have, not sure if that api is worth adding.

**Comment by idoros** [comment]
> unchecked items in the PR description

Like you said, once we agreed on the API, caveats and functionality we can start on documentation or any code/test cleanups

> My reservation is mostly about the usefulness of the closed shadow dom mechanism as it seems to require special access to those components that most devs won't have, not sure if that api is worth adding.

While closed shadow-roots are undetectable, these changes improves the ability of Tabbable within what is possible and offer an additional API in case whoever runs the code does have access to the closed shadow-roots.

Without this code (or another implementation):
- items within shadow-dom are not detected at all
- slotted items are not ordered correctly
- isFocusable/isTabbable has bugs for elements nested in shadow dom (display detection & radio buttons)

And again, I can't tell if it's by design, or if it's a good idea, but if you have access to a page before `customElements.define`,  `attachShadow`, or `attachInternals` are being called, then you can get access and reference all shadow dom and host elements, opened or closed.

**Comment by robrez** [comment]
Hiii, curious of there any updates here?

**Comment by stefcameron** [comment]
I'm still willing to merge it once we've ironed out the issues and have a good feeling it's going to work for those who use the shadow DOM. I don't, so I'm not much help with reviewing. I'd love to have someone run this through its paces to see if we've addressed at least the most glaring corner cases because otherwise, it's going to be really hard for me to support it once it's in the code.

**Comment by idoros** [comment]
@stefcameron would it help if I break it down into smaller PRs?

There are 3 different things here:
- improvement of the internal shadow dom testing/debugging
- small fixes for `isFocusable` and `isTabbable`
- support for querying into shadow dom (that I would do a bit differently now

**Comment by scomea** [comment]
Would a beta package(s) help?  Might make it easier for folks to try it out and give feedback.

**Comment by stefcameron** [comment]
> @stefcameron would it help if I break it down into smaller PRs?
> 
> There are 3 different things here:
> 
> * improvement of the internal shadow dom testing/debugging
> * small fixes for `isFocusable` and `isTabbable`
> * support for querying into shadow dom (that I would do a bit differently now)

Thanks for offering, but it's fine as it is (though it needs rebasing now). I don't think it's complete yet, right?

image

Also, it seemed like you and @scomea were still discussing things. Am I mistaken? All is good?

> Would a beta package(s) help? Might make it easier for folks to try it out and give feedback.

I could publish a beta if folks are ready to look at it.

**Comment by idoros** [comment]
> I don't think it's complete yet, right?

I will finish the docs and changeset once we agree on the API if that's ok

>  it seemed like you and @scomea were still discussing things. Am I mistaken? All is good?

I think this is the best we can get with current closed shadow-dom spec

**Comment by scomea** [comment]
> > I don't think it's complete yet, right?
> 
> I will finish the docs and changeset once we agree on the API if that's ok

I the api bit is the part I'm reserved about as mechanically it seems to work (and revising api's can be painful).   I think having people try it in a beta package could get some more community feedback on that point.

**Comment by stefcameron** [comment]
> > > I don't think it's complete yet, right?
> > 
> > 
> > I will finish the docs and changeset once we agree on the API if that's ok
> 
> I the api bit is the part I'm reserved about as mechanically it seems to work (and revising api's can be painful). I think having people try it in a beta package could get some more community feedback on that point.

OK. I rebased this branch to resolve conflicts and get the latest code. I can publish a beta from this branch, `tabbable@5.3.0-beta.0` but will that be enough?

Will a beta of focus-trap that uses this beta also be needed? If not, what needs to be done in focus-trap in order to properly leverage the new features here?

As I understand it, there's only one change to the tabbable API here, and that's the new
```typescript
getShadowRoot?: (node: FocusableElement) => ShadowRoot | boolean | undefined;
```
option. Will that have to be added to focus-trap in order to pass it down via https://github.com/focus-trap/focus-trap/blob/master/index.js#L228?

stefcameron:
@scomea @idoros Happy New Year! Can you guys help me with [my last comment](https://github.com/focus-trap/tabbable/pull/304#issuecomment-986820582)?

idoros:
Hey, sorry I thought I answered this and somehow my reply is not here 😞.

Yes, exposing getShadowRoot is the only new API at the moment, although we might want to add something for delegatesFocus later on. If there is any one that would test it then a beta would help clear up any issues.

stefcameron:
@idoros 
> Hey, sorry I thought I answered this and somehow my reply is not here 😞.

No worries, thanks for chiming again.

> Yes, exposing getShadowRoot is the only new API at the moment, although we might want to add something for delegatesFocus later on. If there is any one that would test it then a beta would help clear up any issues.

What about `focus-trap`? Do we just need to expose `getShadowRoot()` through the options when setting up a focus-trap?

idoros:
This API should probably be exposed through in order to support closed shadow root. Open roots should just work.

stefcameron:
OK, I will work on making a beta of tabbable with this feature, and making a beta of focus-trap with this beta, and the API exposed.

stefcameron:
I'm going to merge this into a beta branch in the upstream repo to make it easier to publish. We can open a new PR from that branch if we get feedback that things need to be improved.

stefcameron:
FYI, `tabbable@5.3.0-beta.0` has been released with this initial implementation, which is now all on the [beta-530](https://github.com/focus-trap/tabbable/tree/beta-530) branch in `tabbable` and a new PR is open for managing the beta from the repo: https://github.com/focus-trap/tabbable/pull/570

If anyone has fixes or enhancements, please make a PR into that branch.

PR #593, opened 2022-02-25
fix(package.json): A small fix in the "test" script.

DaviDevMod:
In CONTRIBUTING.md there are instructions about some steps to take when contributing to the project. In particular, there is:

>When you're done with your changes, be sure to run `yarn format` to have Prettier format your code, and use `yarn lint` to check for syntax issues. `yarn test:unit` will run unit tests.

>You can also simply run `yarn test` to check all of the above.

However simply running `yarn test` leads to the following output:

```
yarn run v1.22.17
$ yarn format:check && yarn lint && yarn test:types && yarn test:unit && yarn test:e2e
$ prettier --check "{*,src/**/*,test/**/*,.github/workflows/*}.+(js|yml)"
Checking formatting...
[warn] src/index.js
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

And it turns out you actually need to run `yarn format` before `yarn test` because the latter internally runs only `yarn format:check`.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "format:watch": "onchange \"{*,src/**/*,test/**/*,.github/workflows/*}.+(js|yml)\" -- prettier --write {{changed}}",
     "lint": "eslint \"*.js\" \"src/**/*.js\" \"test/**/*.js\"",
     "start": "yarn compile:cjs && budo -l -d -o test/debug.js -- -t brfs",
-    "test": "yarn format:check && yarn lint && yarn test:types && yarn test:unit && yarn test:e2e",
+    "test": "yarn format && yarn lint && yarn test:types && yarn test:unit && yarn test:e2e",
     "test:types": "tsc index.d.ts",
     "test:unit": "jest",
     "test:e2e": "cypress run",


stefcameron:
Thanks for the PR, but the point of the test script is not to modify any code. `test` will run all tests. If you haven't formatted your code yet (sometimes it's easier on the eyes _not_ to format right away, especially if you aren't used to how Prettier formats code), you can run the individual tests like `test:unit` and `lint` separate from checking formatting.

And we run `test` in CI, so if test modifies the code, we won't catch that the code isn't formatted, because the change will happen in the container that runs the tests, and it won't be committed.

DaviDevMod:
Oh, all right. However a script that does what I was expecting from `yarn test` would be helpful too.  
I have prettier on VSC that formats my code on save, but it has a different configuration (eg, it puts a space between the brackets of an empty object: `{ }` rather than `{}`). It would be nice to have a dev environment prettier-configuration-agnostic that automatically applies the standard formatting chosen by the project. It would save a lot of `yarn format`.

PR #597, opened 2022-02-27
fix(index.js): fixed bug in 'isDisabledFromFieldset' (fixes #596)

### LINKED ISSUE #596: Bug: isDisabledFromFieldset doesn't check the top-most disabled <fieldset>
[isDisabledFromFieldset](https://github.com/focus-trap/tabbable/blob/b341412a8650f3c334f38ebda2200ae4c0d84328/src/index.js#L171) is supposed to tell whether a node is in the first `<legend>` of the top-most disabled `<fieldset>`, but is actually assessing whether a node is in the first `<legend>` of the closest ancestor `<fieldset>`, which is not necessarily the top-most `<fieldset>` containing the node.

DaviDevMod:
As noted in #596  `isDisabledFromFieldset` wasn't checking whether the disabled `<fieldset>` containing `node` is the top-most disabled `<fieldset>`. Now it does 🎉

Fixes #596

Note: I have no idea if there are tests for the case in which a form field is nested in the first `<legend>` of a disabled `<fieldset>` being nested in another disabled `<fieldset>`. If there are such tests, they need to be fixed too.

Edit: It looks like there are no such tests. There are no disabled `<fieldset>`s nested in another disabled `<fieldset>` in the structure of [fieldset.html](https://github.com/focus-trap/tabbable/blob/master/test/fixtures/fieldset.html). Should I add such a test?

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/.changeset/long-planets-sleep.md
+++ b/.changeset/long-planets-sleep.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+Fixed bug in `isDisabledFromFieldset`. The function wasn't checking whether the disabled <fieldset> containing `node` is the top-most disabled <fieldset>.
--- a/.changeset/long-planets-sleep.md
+++ b/.changeset/long-planets-sleep.md
@@ -2,4 +2,4 @@
 'tabbable': patch
 ---
 
-Fixed bug in `isDisabledFromFieldset`. The function wasn't checking whether the disabled <fieldset> containing `node` is the top-most disabled <fieldset>.
+Fixed bug in `isDisabledFromFieldset`. The function wasn't checking whether the disabled `<fieldset>` containing `node` is the top-most disabled `<fieldset>` ([#596](https://github.com/focus-trap/tabbable/issues/596)).
--- a/src/index.js
+++ b/src/index.js
@@ -166,37 +166,35 @@ const isHidden = function (node, displayCheck) {
 };
 
 // form fields (nested) inside a disabled fieldset are not focusable/tabbable
-// unless they are in the _first_ <legend> element of the top-most disabled fieldset
+//  unless they are in the _first_ <legend> element of the top-most disabled
+//  fieldset
 const isDisabledFromFieldset = function (node) {
   if (/^(INPUT|BUTTON|SELECT|TEXTAREA)$/.test(node.tagName)) {
-    let parentNode = node;
-    let disabledFieldsetsCount = 0;
-    let legend;
-    // count how many disabled <fieldset> (DF) are ancestors of `node`
-    while ((parentNode = parentNode.parentElement)) {
+    let parentNode = node.parentElement;
+    // check if `node` is contained in a disabled <fieldset>
+    while (parentNode) {
       if (parentNode.tagName === 'FIELDSET' && parentNode.disabled) {
-        // if `node` has more than one DF ancestor, then it is not focusable
-        if (++disabledFieldsetsCount === 2) {
-          return true;
-        }
-        // find the first legend in the closest DF ancestor of `node`
+        // look for the first <legend> among the children of the disabled <fieldset>
         for (let i = 0; i < parentNode.children.length; i++) {
-          if (parentNode.children.item(i).tagName === 'LEGEND') {
-            legend = parentNode.children.item(i);
-            break;
+          const child = parentNode.children.item(i);
+          // when the first <legend> (in document order) is found
+          if (child.tagName === 'LEGEND') {
+            // if its parent <fieldset> is not nested in another disabled <fieldset>,
+            // return whether `node` is a descendant of its first <legend>
+            return matches.call(parentNode, 'fieldset[disabled] *')
+              ? true
+              : !child.contains(node);
           }
         }
-        if (!legend) {
-          // if the closest DF ancestor has no <legend>, then `node` is not focusable
-          return true;
-        }
+        // the disabled <fieldset> containing `node` has no <legend>
+        return true;
       }
+      parentNode = parentNode.parentElement;
     }
-    // if exactly one DF ancestor was found, return whether its legend contains `node`;
-    // if no DF ancestor was found, return `false` (meaning `node` is focusable).
-    return disabledFieldsetsCount ? !legend.contains(node) : false;
   }
-  // `node` is not a form field and its focusability can't be assessed here.
+
+  // else, node's tabbable/focusable state should not be affected by a fieldset's
+  //  enabled/disabled state
   return false;
 };
 
--- a/src/index.js
+++ b/src/index.js
@@ -173,19 +173,26 @@ const isDisabledFromFieldset = function (node) {
     let parentNode = node.parentElement;
     while (parentNode) {
       if (parentNode.tagName === 'FIELDSET' && parentNode.disabled) {
-        // look for the first <legend> as an immediate child of the disabled
-        //  <fieldset>: if the node is in that legend, it'll be enabled even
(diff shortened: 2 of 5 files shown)


DaviDevMod:
I added a test for the edge case in question.

DaviDevMod:
I did run `yarn format` before to commit, not sure why the formatting has changed.

DaviDevMod:
I added a test for `isFocusable`. previously I only edited `fieldset.html` which (for what I understand) is being picked up by `focusable.e2e` and `tabbable.e2e`.

Now I provided a test for `isFocusable.e2e` and that's great, but I think I could have done more, though I can't because I am facing this error:

> TestingLibraryElementError: Unable to find an element by: [data-testid="fieldset-disabled-fieldset-disabled-legend-button"]

I don't know why the element is not found, and I guess it's because I don't know cypress. So I leave further testing up to someone else. [Here is a branch](https://github.com/DaviDevMod/tabbable/tree/test/isFocusable) in which you can reproduce the error by running `yarn test`.

DaviDevMod:
For what I understand, the changes I made to `fieldset.html` will be picked up only by `focusable.e2e` and `tabbable.e2e`.

I tried to add a test to `isFocusable.e2e` and am facing this error:

> TestingLibraryElementError: Unable to find an element by: [data-testid="fieldset-disabled-fieldset-disabled-legend-button"]

I think it's time to leave the work to someone more knowledgeable about cypress. [Here is a branch](https://github.com/DaviDevMod/tabbable/tree/test/isFocusable) on my fork where you can reproduce the error by running `yarn test`.

DaviDevMod:
Since once a child `<legend>` is found, the function will return in any case, it is possible to use `parentNode` in the inner while loop without messing with the outer loop. So I removed the unnecessary `ancestor` variable and also added more comments, because the function's logic is tricky and not so easy to follow.

test/fixtures/fieldset.html
```diff
@@ -60,6 +60,11 @@
     <input id="fieldset-disabled-fieldset-enabled-input" type="text" value="Input not tabbable/focusable" />
   </fieldset>
 
+  <fieldset id="fieldset-disabled-fieldset-disabled" disabled>
+    <legend><button id="fieldset-disabled-fieldset-disabled-legend-button">Button not tabbable/focusable</button></legend>
+    <input id="fieldset-disabled-fieldset-disabled-input" type="text" value="Input not tabbable/focusable" />
+  </fieldset>
```
stefcameron, line 66: Thanks for adding this! The old version of the code doesn't catch this one. 🎉 

Adding this covers the following tests:
- https://github.com/focus-trap/tabbable/blob/master/test/e2e/tabbable.e2e.js#L259
- https://github.com/focus-trap/tabbable/blob/master/test/e2e/focusable.e2e.js#L230

But it doesn't cover these two:
- https://github.com/focus-trap/tabbable/blob/master/test/e2e/isFocusable.e2e.js#L302
- https://github.com/focus-trap/tabbable/blob/master/test/e2e/isTabbable.e2e.js#L293

For some reason, those last 2 test files don't use the fixtures. 🤷‍♂️ 

Can you add to those so that `isFocusable()` and `isTabbable()` are tested for this additional case you uncovered that we weren't handling properly?

src/index.js
```diff
@@ -171,24 +171,31 @@ const isHidden = function (node, displayCheck) {
 const isDisabledFromFieldset = function (node) {
   if (/^(INPUT|BUTTON|SELECT|TEXTAREA)$/.test(node.tagName)) {
     let parentNode = node.parentElement;
+    // check if `node` is contained in a disabled <fieldset>
     while (parentNode) {
+      // if it is
```
stefcameron, line 176: I think this comment (and the other identical one at line 185) could be dropped. They're a bit "over the top", IMO. The others are helpful. Thank you! That will help with posterity.

.changeset/long-planets-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+Fixed bug in `isDisabledFromFieldset`. The function wasn't checking whether the disabled <fieldset> containing `node` is the top-most disabled <fieldset>.
```
stefcameron, line 5: Please put `<fieldset>` into backticks so it's rendered as code. Otherwise, Markdown accepts HTML and an open `<fieldset>` tag in there will mess-up the CHANGELOG.

Also, I like to reference the issue the fix is related to at the end, like this: `([#596](https://github.com/focus-trap/tabbable/issues/596))` (you have to link it because this is going into the CHANGELOG, which isn't a PR or an Issue, so just `#596` won't result in an auto-link from GitHub).

stefcameron (changes requested):
Nice work! I'd just like to cover `isFocusable()` and `isTabbable()` with an additional use case to check for what you discovered.

stefcameron:
> Edit: It looks like there are no such tests. There are no disabled <fieldset>s nested in another disabled <fieldset> in the structure of [fieldset.html](https://github.com/focus-trap/tabbable/blob/master/test/fixtures/fieldset.html). Should I add such a test?

I think you have now, right? That's what I saw in your PR so far.

DaviDevMod:
I think you have now, right? That's what I saw in your PR so far.

Yes I covered the case by adding a nested disabled `<feldset>` in `fieldset.html`. But if I add the same `<fieldset>` in the HTML present in `isFocusable.e2e` and `isTabbable.e2e`, then I get an error that at the moment I don't know how to overcome.

Anyway I am about to start a long shift at work and I won't be able to touch GitHub for at least 10 hours from now. Just to let you know, eventually you will hear from me again.

stefcameron:
Don't sweat it! Whenever you have time again. I appreciate your help on the project.

DaviDevMod:
I just added a commit with a different logic that counts the number of disabled `fieldset` being ancestors of `node`, in this way the inner and outer `while` loops can be merged together. Let me know which one is the preferred approach here.

DaviDevMod:
I just committed a more compact version of the original logic. Here the inner while loop has been replaced by

```javascript
matches.call(parentNode, 'fieldset[disabled] *')
```

For some unknown reason, the coverage is slightly different for the three approaches:
- the original approach shows **+0.05%**
- the one with a different logic shows **+0.07%** (and a line not covered, so potential to increment coverage)
- the approach I just committed shows **+0.03%**

But I guess it's not that relevant.

By the way, I apologise if I am making this PR somehow more complex than it is supposed to be.

src/index.js
```diff
@@ -171,24 +171,24 @@ const isHidden = function (node, displayCheck) {
...
           const child = parentNode.children.item(i);
+          // when the first <legend> (in document order) is found
           if (child.tagName === 'LEGEND') {
-            // return whether `node` is a descendant of the first <legend> found
-            return !child.contains(node);
+            // if its parent <fieldset> is not nested in another disabled <fieldset>,
+            // return whether `node` is a descendant of its first <legend>
+            return matches.call(parentNode, 'fieldset[disabled] *')
+              ? true
+              : !child.contains(node);
```
stefcameron, line 186: Smart. I like this better than the inner while loop from the first iteration, and it's cleaner than your second iteration.

stefcameron (approved):
This looks good to me! Thank you once again! And also thanks for cleaning-up the fixtures in those two test modules.

DaviDevMod:
> also thanks for cleaning-up the fixtures in those two test modules.

Being able to do some work on preexisting tests, without having to care about all the boilerplate, is making learning cypress much more enjoyable. So thank you!

PR #604, opened 2022-03-02
Optimizing `displayCheck: "full"`

### LINKED ISSUE #592: Optimizing displayCheck: 'full'
It is possible to assess whether an element, or one of its ancestors, have `display: none` without having to traverse the DOM.

[This condition](https://github.com/focus-trap/tabbable/blob/7609e9ecf47941a2b69aab2149e8760ea1f7392e/src/index.js#L153) can be simplified to:

```javascript
if (!displayCheck || displayCheck === 'full') {
  return !node.getClientRects().length;
}
```

I actually don't know if checking for `getClientRects().length` is really equivalent to that expensive `while` loop.
 
According to the specs, the [`getClientRects()`](https://drafts.csswg.org/cssom-view/#dom-element-getclientrects) method returns an empty list

> If the element on which it was invoked does not have an associated [layout box](https://drafts.csswg.org/cssom-view/#layout-box)

What exactly are all the cases in which an element does not have an associated layout box? I am not sure.

But the e2e tests currently present in the project are all passing.  
However I didn't go through the tests myself and I don't know what they do, I only checked [this very small set of possible edge cases](https://codesandbox.io/s/great-frog-65yr2y?file=/index.html) and it looks like the method should be good enough.

May I open a PR?

By the way, since the [`getBoundingClientRect()`](https://drafts.csswg.org/cssom-view/#dom-element-getboundingclientrect) method internally uses `getClientRects()`, the `"full"` option would become faster than the `"non-zero-area"` one. But I guess that the latter would still be useful for accessibility purpose.

DaviDevMod:
Optimized `displayCheck: "full"` by replacing [this DOM traversal](https://github.com/focus-trap/tabbable/blob/7609e9ecf47941a2b69aab2149e8760ea1f7392e/src/index.js#L154) (executed on every potential focusable element) with the following statement:

```javascript
return !node.getClientRects().length;
```

which is a more general check for whether an elements has display boxes. `display: "none"` is only a special case of `!node.getClientRects().length`.

Another case of elements with no display boxes would be elements with `display: "contents"`. And (since this case wasn't previosly handled) tests have been added to cover this possibility.

Fixes #592

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/.changeset/hip-tomatoes-hug.md
+++ b/.changeset/hip-tomatoes-hug.md
@@ -1,0 +1,5 @@
+---
+'tabbable': patch
+---
+
+Optimized and extended `displayCheck: "full"` option (now checks for any element having no display boxes) and added test for `display: "contents"` property (this bug was never reported). [(#592)](https://github.com/focus-trap/tabbable/issues/592)
--- a/src/index.js
+++ b/src/index.js
@@ -151,12 +151,7 @@ const isHidden = function (node, displayCheck) {
     return true;
   }
   if (!displayCheck || displayCheck === 'full') {
-    while (node) {
-      if (getComputedStyle(node).display === 'none') {
-        return true;
-      }
-      node = node.parentElement;
-    }
+    return !node.getClientRects().length;
   } else if (displayCheck === 'non-zero-area') {
     const { width, height } = node.getBoundingClientRect();
     return width === 0 && height === 0;
--- a/test/e2e/focusable.e2e.js
+++ b/test/e2e/focusable.e2e.js
@@ -330,6 +330,7 @@ describe('focusable', () => {
             'displayed-top',
             'displayed-nested',
             'displayed-zero-size',
+            'nested-under-displayed-contents',
           ];
           const container = document.createElement('div');
           container.innerHTML = fixtures.displayed;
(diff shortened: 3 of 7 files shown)


DaviDevMod:
Not sure why the coverage has decreased. It may be that there are cases other than the `display: "full"` and `display: "contents"` ones, in which an element has no display boxes.

stefcameron:
> Not sure why the coverage has decreased. It may be that there are cases other than the `display: "full"` and `display: "contents"` ones, in which an element has no display boxes. And we have no tests for these additional cases.

I think the reason it has decreased is because technically, there were 4 lines that were covered before, and now there is only one:

image

But this is a blind comparison between before (say 300 lines total) and now (say 296 lines total), so technically, coverage has _decreased_ by a fraction, but that's not because there are lines that were covered before and aren't now. It's because there are just less lines covered.

Hopefully that makes sense. Not something to worry about in this case.

stefcameron (approved):
This looks good, thank you!

DaviDevMod:
> I think the reason it has decreased is because technically, there were 4 lines that were covered before, and now there is only one

Oh so that's how the coverage is calculated.  
I don't know what kind of smarter calculation I was expecting. But yes, recognising which lines are covered by tests, is already pretty clever.

Thank you for the clarification!

stefcameron:
@all-contributors add @DaviDevMod for code, test

craigkovatch:
This broke ~all of our unit tests which touch code that uses `tabbable`, opening an issue...

PR #610, opened 2022-03-11
fix(index.js) The tabIndex of contentEditable elements was assumed to…

DaviDevMod:
… be zero in any case, not only in the case it was not specifically set.

Apart from fixing the little bug, I took the freedom to remove the `isContentEditable` function because it consists of a single statement and is being used only once. Also used a regex for checking `node.nodeName`.  
I can revert these changes if they are undesired.

Also I think that the logic of `getTabIndex` [can be simplified](https://github.com/DaviDevMod/tabbable/commit/d7a987be47270d980dde353f4a024d3fc6621a6e). Let me know if you want these changes in this current PR.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -31,10 +31,6 @@ const getCandidates = function (el, includeContainer, filter) {
   return candidates;
 };
 
-const isContentEditable = function (node) {
-  return node.contentEditable === 'true';
-};
-
 const getTabindex = function (node) {
   const tabindexAttr = parseInt(node.getAttribute('tabindex'), 10);
 
@@ -42,21 +38,15 @@ const getTabindex = function (node) {
     return tabindexAttr;
   }
 
-  // Browsers do not return `tabIndex` correctly for contentEditable nodes;
-  // so if they don't have a tabindex attribute specifically set, assume it's 0.
-  if (isContentEditable(node)) {
-    return 0;
-  }
-
   // in Chrome, <details/>, <audio controls/> and <video controls/> elements get a default
-  //  `tabIndex` of -1 when the 'tabindex' attribute isn't specified in the DOM,
-  //  yet they are still part of the regular tab order; in FF, they get a default
-  //  `tabIndex` of 0; since Chrome still puts those elements in the regular tab
-  //  order, consider their tab index to be 0.
+  // `tabIndex` of -1 when the 'tabindex' attribute isn't specified in the DOM,
+  // yet they are still part of the regular tab order; in FF, they get a default
+  // `tabIndex` of 0; since Chrome still puts those elements in the regular tab
+  // order, consider their tab index to be 0.
+  // Also browsers do not return `tabIndex` correctly for contentEditable nodes;
+  // so if they don't have a tabindex attribute specifically set, assume it's 0.
   if (
-    (node.nodeName === 'AUDIO' ||
-      node.nodeName === 'VIDEO' ||
-      node.nodeName === 'DETAILS') &&
+    (/^(AUDIO|VIDEO|DETAILS)$/.test(node.tagName) || node.isContentEditable) &&
     node.getAttribute('tabindex') === null
   ) {
     return 0;


stefcameron (changes requested):
Thanks for noticing that bug! Could you please rebase this change on top of the `beta-530` branch (so make a PR into that branch instead of `master`), add a changeset, and also add a test case to cover this bug?

stefcameron:
> Also I think that the logic of `getTabIndex` [can be simplified](https://github.com/DaviDevMod/tabbable/commit/d7a987be47270d980dde353f4a024d3fc6621a6e). Let me know if you want these changes in this current PR.

Thanks for the suggestion, but I think this change would be considering a node that is content editable with a set negative tab index (so focusable but not tabbable) to actually be tabbable, wouldn't it?

DaviDevMod:
> Thanks for noticing that bug! Could you please rebase this change on top of the beta-530 branch (so make a PR into that branch instead of master), add a changeset, and also add a test case to cover this bug?

Of course. Regarding making making a PR into `beta-530`, you mean to close this PR and open a brand new one right? Just want to double check.

> Thanks for the suggestion, but I think this change would be considering a node that is content editable with a set negative tab index (so focusable but not tabbable) to actually be tabbable, wouldn't it?

No `0` is returned only if `&& node.getAttribute('tabindex') === null`, so if the tab index was set to `-1` (or anything) the condition would not be entered and the function will return `node.tabIndex`.

Maybe the outer condition `if (node.tabIndex < 0)` generated some confusion. This condition is actually unnecessary and can be removed.
The fact is that browsers can't default to positive tab indexes, messing with the tab order, so we know that if `node.tabindex` is positive, then `&& node.getAttribute('tabindex') === null` must return `false`.  
If instead `node.tabindex` is `0` (no matter if it comes from the browser default or if it was explicitly set), then there is no need to go through the check just to `return 0;`  
So the only case in which the check is actually needed is when `node.tabIndex` is negative.  
But, of course, we can remove the outer condition and run the inner check in any case.

[Just for completeness](https://html.spec.whatwg.org/multipage/interaction.html#dom-tabindex): `.tabIndex` is granted to be an integer and its default must be either 0 or -1.

DaviDevMod:
I looked at where `getTabIndex` is being used in order to understand whether the outer condition `if (node.tabIndex < 0)` discussed above is good for performance, and indeed it is. The reason is that the `node` parameter comes from `getCandidate`, thus it is very likely that its `tabIndex` is 0 (given by the browser as default).

Also, I noticed something else that can do its bit to improve performance. In this function:

```javascript
const isNodeMatchingSelectorTabbable = function (options, node) {
  if (
    !isNodeMatchingSelectorFocusable(options, node) ||
    isNonTabbableRadio(node) ||
    getTabindex(node) < 0
  ) {
    return false;
  }
  return true;
};
```
since `isNodeMatchingSelectorFocusable` is quite expensive, it would be nice to reduce the chances to call it, by simply changing the order of the function calls:

```javascript
  if (
    isNonTabbableRadio(node) ||
    getTabindex(node) < 0 ||
    !isNodeMatchingSelectorFocusable(options, node)
  ) 
```
If you want I can add this simple line swapping to this PR.

stefcameron:
> Of course. Regarding making making a PR into beta-530, you mean to close this PR and open a brand new one right? Just want to double check.

Thanks! You should be able to change the destination branch of this PR to be `beta-530`, and then locally, you can create a new branch based on `beta-530`, cherry-pick your commit for this fix onto it, and force-push that up to this PR branch. At least that's what I would do. If it's easier for you to start with a new PR, though, you can close this one and make a new one.

> Also, I noticed something else that can do its bit to improve performance.

Yes, making better use of short-circuit evaluation there would be a good performance gain!

stefcameron:
In case you didn't know, when you edit your PR title in GitHub, you also get the ability to change the destination branch:

Screen Shot 2022-03-12 at 3 08 36 PM

So that or a new PR, either way is fine. If you re-use this one, you'll just need to rebase it onto `beta-530` locally and force-push the commits up (after changing the PR's destination branch in GitHub).

stefcameron:
If you could also add an entry in the `CHANGELOG.md` file under a new `## UNRELEASED` heading as a bullet point with the same text you'll use in your Changeset, that would be great.

If I publish another beta, I'll know to mention this change in it. I still want the Changeset since it'll get used when the `beta-530` branch gets merged to master and the official v5.3.0 is published.

DaviDevMod:
> Thanks! You should be able to change the destination branch of this PR to be beta-530, and then locally, you can create a new branch based on beta-530, cherry-pick your commit for this fix onto it, and force-push that up to this PR branch.

Thank you so much for the guide. I am going to do it.

> In case you didn't know, when you edit your PR title in GitHub, you also get the ability to change the destination branch

You got me, I didn't know. Thanks again!

> If you could also add an entry in the CHANGELOG.md file under a new ## UNRELEASED heading as a bullet point with the same text you'll use in your Changeset, that would be great.

Absolutely.

DaviDevMod:
The rebase worked, the PR automatically picked up the force-pushed commits. :tada:

Now as you can see there are seven tests not passing. However I think that they are actually passing, but cypress in not smart or configured enough to realise it:

Screenshot from 2022-03-13 18-31-13

All and only the expected elements are found, but they appear in a different order in the array..  
Not sure what caused this difference. May it be one of the dependabot version bumps?

DaviDevMod:
It is possible to make the tests pass by using [`to.have.members`](https://www.chaijs.com/api/bdd/#method_members) rather `to.eql` when asserting the result. However for some of the tests the ordering of the elements actually matters:

Screenshot from 2022-03-13 19-28-42

I don't have any knowledge about shadow DOMs, but I will have a look at the code being tested nevertheless. Just to be sure the bug isn't in the tests, before to look for it in `src`.

stefcameron:
I just spent a bunch of time this past week merging #604 into this branch, scratching my head about the Shadow DOM stuff. 😉 It's someone else's (nice) contribution with which I'm increasingly getting familiar with now that I'm having to support it... I knew zero about shadow DOM myself, and still a noob there. It's not something I use on a daily basis, or have every used.

In the end, the order of the nodes matters, and the order is dictated by the IF statement here: https://github.com/focus-trap/tabbable/blob/beta-530/src/index.js#L113

See the comment there where I attempt to explain to my future self what on earth is going on there. 😆 But basically, when shadow support is enabled, if there's an undisclosed shadow (i.e. shadow root is closed, but `getShadowRoot()` __returns__ `true`), we attempt to _more closely_ mimic the tab order you would get had you disclosed it, and we do this by treating the element that has the undisclosed shadow as a kind of parent by recursively checking its children and then creating a scope here: https://github.com/focus-trap/tabbable/blob/beta-530/src/index.js#L129

All that to say, the resulting order of the nodes will be slightly different between `getShadowRoot: false` or `getShadowRoot: true`, and `getShadowRoot: () => true`. The idea is that the order of the nodes with `getShadowRoot: () => true` would be _somewhat closer_ to what you would get if you had `getShadowRoot: (node) => closedShadowRoot` which would tell tabbable about the shadow and let it recurse its children.

Why? I'm not 100% sure, but people who have been testing the beta with web components so far seem happy, so I'm not inclined to remove this logic right now.

LMK if there's something else I can try to explain.

Strange that changing how we check the tab order has affect order here, but I'm hoping it's the tab order change that has broken things since that would be logical (though still as yet a mystery). As of my work yesterday, all the tests are passing.

DaviDevMod:
I found that the `beta-530` branch has `getTabIndex` getting two parameters, rather than one. This was the bug.  

> LMK if there's something else I can try to explain.

Thank you for your kind availability :pray: I do have a question for you.

Since (as you can see from this [little example](https://codesandbox.io/s/affectionate-curie-m9qtev?file=/index.html)) the DOM is queried and tabbed in [depth-first order](https://en.wikipedia.org/wiki/Depth-first_search).

I suppose that shadow DOMs works with the same order. Once you replace slots and anything (as I said I don't know about shadow DOM) you are left with a regular DOM tree which most probably will be tabbed in depth-first order by the browser.

Since there is a specific and unambiguous order to follow, why can't it be followed in any case? (whatever `getShadowRoot` is or returns)

DaviDevMod:
I wrote tests for audio and video elements (just because `contenteditable`s were already handled properly), also added changeset and modified CHANGELOG.md

stefcameron:
> Thank you for your kind availability 🙏 I do have a question for you.

My pleasure!

> Since (as you can see from this [little example](https://codesandbox.io/s/affectionate-curie-m9qtev?file=/index.html)) the DOM is queried and tabbed in [depth-first order](https://en.wikipedia.org/wiki/Depth-first_search).
> 
> I suppose that shadow DOMs work with the same order. Once you replace slots and anything (as I said I don't know about shadow DOM) you are left with a regular DOM tree which most probably will be tabbed in depth-first order by the browser.
> 
> Since there is a specific and unambiguous order to follow, why can't it be followed in any case? (whatever `getShadowRoot` is or returns)
> 
> Edit: in your comment above there was
> 
> > Why? I'm not 100% sure, but people who have been testing the beta with web components so far seem happy, so I'm not inclined to remove this logic right now.
> 
> Which more or less answers my question 😅 I guess it's fine to ship with this logic, if someone ever complaints, then it can always be readjusted later.

I get your thinking, but yes, in this case, I'm inclined to go with the flow and adjust later if it's really wrong.

stefcameron:
> I wrote tests for audio and video elements (just because `contenteditable`s were already handled properly), also added changeset and modified CHANGELOG.md

Thank you! Much appreciated!

stefcameron (approved):
This all looks good to me! And I really appreciate the additional tests.

stefcameron:
I'm just going to see if I can remove the package.json and yarn.lock updates from this PR since they shouldn't affect the outcome. I'm guessing those came in from maybe a rebase since I think those changes are on master. I'll rebase the beta-530 branch separately after we merge this.

DaviDevMod:
I learned a lot of things in this PR, the most important one being: *never report a bug before having written a failing test* :laughing:

Thank you for the patience and all the explanations! :rocket:

stefcameron:
👍 👍

PR #653, opened 2022-04-28
Update docs for `displayCheck` configuration.

DaviDevMod:
Hello there!

While passing by the repo I noticed an outdated documentation for `displayCheck` and I felt responsible for it :sweat_smile:

I am not sure about this line I'm introducing:

> it also used to be less intensive than the `full` option, but as a result of optimizations this is not the case anymore.

It is meant to help justifying the existence of the `non-zero-area` option, but it's not a necessary information and may as well be omitted.

I also noticed you landed the Shadow DOM feature :rocket: keep up the good work!

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ To reliably check if an element is tabbable/focusable, Tabbable defaults to the 
 
 The `displayCheck` configuration accepts the following options:
 
-- `full`: (default) Most reliably resembling browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (notice that this doesn't exclude `visibility: hidden` or elements with zero size). This check is by far the slowest option as it will cause layout reflow.
-- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option may be less intensive than the `full` option, and better for accessibility, as zero-size elements with focusable content are considered a strong accessibility anti-pattern.
+- `full`: (default) Most reliably resembling browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (notice that this doesn't exclude `visibility: hidden` or elements with zero size).
+- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option may enhance accessibility, as zero-size elements with focusable content are considered a strong accessibility anti-pattern; it also used to be less intensive than the `full` option, but as a result of optimizations this is not the case anymore.
 - `none`: This completely opts out of the display check. **This option is not recommended**, as it might return elements that are not displayed, and as such not tabbable/focusable and can break accessibility. Make sure you know which elements in your DOM are not displayed and can filter them out yourself before using this option.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -137,8 +137,8 @@ To reliably check if an element is tabbable/focusable, Tabbable defaults to the
 
 The `displayCheck` configuration accepts the following options:
 
-- `full`: (default) Most reliably resembling browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (notice that this doesn't exclude `visibility: hidden` or elements with zero size). This check is by far the slowest option as it will cause layout reflow.
-- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option may be less intensive than the `full` option, and better for accessibility, as zero-size elements with focusable content are considered a strong accessibility anti-pattern.
+- `full`: (default) Most reliably resembling browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (notice that this doesn't exclude `visibility: hidden` or elements with zero size).
```
stefcameron, line 140: I think it's still worth pointing out that this option causes layout reflow, since that was a big point of contention in the past, which led to these 3 `displayCheck` options (where we only had the "full" in the past).

```suggestion
- `full`: (default) Most reliably resembling browser behavior, this option checks that an element is displayed and all of his ancestors are displayed as well (notice that this doesn't exclude `visibility: hidden` or elements with zero size). This option will cause layout reflow, however. If that is a concern, consider the `none` option.
```

README.md
```diff
@@ -137,8 +137,8 @@ To reliably check if an element is tabbable/focusable, Tabbable defaults to the
...
+- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option may enhance accessibility, as zero-size elements with focusable content are considered a strong accessibility anti-pattern; it also used to be less intensive than the `full` option, but as a result of optimizations this is not the case anymore.
```
stefcameron, line 141:
```suggestion
- `non-zero-area`: This option checks display under the assumption that elements that are not displayed have zero area (width AND height equals zero). While not keeping true to browser behavior, this option may enhance accessibility, as zero-size elements with focusable content are considered a strong accessibility anti-pattern. Like the `full` option, this option also causes layout reflow, and should have basically the same performance. Consider the `none` option if reflow is a concern.
```

stefcameron (commented):
Thanks for noticing and offering a fix! I've included my suggestions. WDYT?

stefcameron:
> I also noticed you landed the Shadow DOM feature 🚀 keep up the good work!

Thank you! Glad to have it finally done. 😄

README.md
DaviDevMod: > I think it's still worth pointing out that this option causes layout reflow, since that was a big point of contention in the past

Absolutely!

stefcameron (approved):

stefcameron:
Thank you, @DaviDevMod !

stefcameron:
@all-contributors add @DaviDevMod for docs

PR #681, opened 2022-05-19
fix: tabbable sometimes incorrectly includes children of a web component

benjamin-t-frost:
## Use Case
A browser will skip tab-targeting an element if it is the only child of a web component which has a tabindex of -1.

## Example

```html
<my-input tabindex="-1">
  <input id="input" type="text" />
</my-nput>
```

## Expected behaviour
`tabbable()` should return an empty list for the example above

## Actual behaviour
`tabbable()` returns a list which includes the input depicted above

## Impact
This bug can cause a focus-trap to behave erratically when nearing the end of focusable elements.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -50,6 +50,12 @@ const getCandidates = function (el, includeContainer, filter) {
  */
 
 /**
+ * @callback ShadowRootFilter
+ * @param {Element} shadowHostNode the element which contains shadow content
+ * @returns {boolean} true if a shadow root could potentially contain valid candidates.
+ */
+
+/**
  * @typedef {Object} CandidatesScope
  * @property {Element} scope contains inner candidates
  * @property {Element[]} candidates
@@ -62,6 +68,7 @@ const getCandidates = function (el, includeContainer, filter) {
  *  or a boolean stating if it has an undisclosed shadow root
  * @property {(node: Element) => boolean} filter filter candidates
  * @property {boolean} flatten if true then result will flatten any CandidatesScope into the returned list
+ * @property {ShadowRootFilter} shadowRootFilter filter shadow roots;
  */
 
 /**
@@ -110,7 +117,10 @@ const getCandidatesIteratively = function (
         (typeof options.getShadowRoot === 'function' &&
           options.getShadowRoot(element));
 
-      if (shadowRoot) {
+      const validShadowRoot =
+        !options.shadowRootFilter || options.shadowRootFilter(element);
+
+      if (shadowRoot && validShadowRoot) {
         // add shadow dom scope IIF a shadow root node was given; otherwise, an undisclosed
         //  shadow exists, so look at light dom children as fallback BUT create a scope for any
         //  child candidates found because they're likely slotted elements (elements that are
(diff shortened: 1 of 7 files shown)


stefcameron:
👀

src/index.js
```diff
@@ -415,6 +420,15 @@ const isNodeMatchingSelectorTabbable = function (options, node) {
   return true;
 };
 
+const isValidShadowCandidatesTabbable = function (node, nestedCandidates) {
```
stefcameron, line 423: So here, `node` is expected to be the shadow host node, and `nestedCandidates` are all the candidate tabbable/focusable (already vetted as such) nodes that were found inside of it?

If so, could we clarify the expectation by renaming `node` to `shadowHostNode`?

test/fixtures/shadow-dom-untabbable.html
```diff
@@ -0,0 +1,7 @@
+<test-shadow tabindex="-1">
+  <template shadowroot="open">
+    <div id="container">
+      <input id="input" type="text" />
+    </div>
+  </template>
+</test-shadow>
```
stefcameron, line 7: I appreciate the new test, thank you! Could we also have another test that shows when the `nestedCandidates.length !== -1` case, that is, a test that checks the opposite of what this new test is checking?

Based on your explanation in the code:
```
  // If a custom element with a tabindex of -1 contains a single otherwise tabbable
  // element, browsers will not allow tab targeting said child.
  return nestedCandidates.length !== 1;
```
I'm getting the impression that if I changed your test to:
```
<test-shadow tabindex="-1">
  <template shadowroot="open">
    <div id="container">
      <input id="input" type="text" />
      <input id="input2" type="text" />
    </div>
  </template>
</test-shadow>
```
that now, both inputs would be tabbable? Seems a bit strange, so I wanted to double-check. Having a passing test that checks for the opposite is always a good reminder of all the corner cases.

stefcameron (changes requested):
A very proactive PR, ticking all the boxes. 😄 Thank you! I do have a couple of requests/concerns, though.

src/index.js
```diff
@@ -415,6 +420,15 @@ const isNodeMatchingSelectorTabbable = function (options, node) {
...
+  if (node.getAttribute('tabindex') !== '-1') {
```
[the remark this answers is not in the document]
benjamin-t-frost: After testing, I've changed this to filter out anything which is explicitly negative. `getTabindex()` returns implicit `-1`s and causes things to be filtered out which should not.

test/fixtures/shadow-dom-untabbable.html
stefcameron: Definitely! Thanks for checking it out.

stefcameron (approved):
Thanks for the updates, and for tweaking the comment in the code about what is and is not tabbable to be more generic so there's no confusion in the future about shadows with 1 vs multiple children.

Looks good!

stefcameron:
`focus-trap@6.9.3` is getting published now. :)

Issue #664, opened 2022-05-05
v5.3.0 breaks literally all of our usage of `tabbable` 😅 when called on a node not attached to the document


craigkovatch:
The change to use `getClientRects().length` completely breaks this library for us: https://github.com/focus-trap/tabbable/pull/604

We use `tabbable` when initializing dialogs, to find candidate element(s) to auto-focus when the dialog appears. These dialogs are React components and the call to `tabbable()` occurs during the `componentDidMount` lifecycle method. At that moment, the DOM elements we want to traverse using `tabbable` are fully formed, but they are not yet attached to the document. So `getClientRects().length` always returns 0, and thus every single one of our calls to `tabbable` fails. (At this library level, we have no way to control when the dialog is actually attached to the document.)

Could this change be reverted, or can we scope it to only be used when `document.body.contains(node)`?

craigkovatch:
As some additional context, this broke out from under us, as a result of declaring our dependency on tabbable as `^5.2.1` rather than pinned `5.2.1`. I expect there are plenty of other projects which are also going to be broken in the same way -- next time they regenerate their lock files or upgrade a dependency which depends on `tabbable`, they will inherit this bug without changing anything or knowing why it suddenly broke.

stefcameron:
Hey, thanks for reaching out! Just wondering, are you using focus-trap-react, or are you using tabbable on its own with the sole purpose of identifying focusable nodes before your dialog is rendered?

Also, how is the node not in the tree yet? The definition of `componentDidMount` is "componentDidMount() is invoked immediately after a component is mounted (inserted into the tree)." Refs should be available at this point as well, which should imply the node has been attached to the DOM tree.

> (At this library level, we have no way to control when the dialog is actually attached to the document.)

But if you're not in control of this dialog (I assume, based on this, that it's a third-party library you're using, along with tabbable), so somehow it's still not attached to the DOM when `componentDidUpdate` fires (which, by React's definition of the API, it _should_ be in the DOM), have you tried configuring tabbable with `displayCheck: "none"`?

That configuration would not rely on `getClientRects()`, but it also does not discriminate in any way on what is visible vs hidden.

Does this dialog library have a hook so you can tell when it's been attached to the DOM?

> I expect there are plenty of other projects which are also going to be broken in the same way -- next time they regenerate their lock files or upgrade a dependency which depends on tabbable, they will inherit this bug without changing anything or knowing why it suddenly broke.

[...]

PR #764, opened 2022-08-10
[BREAKING][#666] Default full to attached nodes only; add legacy-full

### LINKED ISSUE #666: [MAJOR] isHidden should exclude detached nodes by default
Change next major version to swap the default behavior of `displayCheck='full'/undefined` to reflect the fact that detached nodes are not actually tabbable/focusable, with an option to ignore `isHidden` on detached nodes.

_Originally posted by @idoros in https://github.com/focus-trap/tabbable/issues/664#issuecomment-1119941242_

stefcameron:
Fixes #666

Tabbable has long had legacy behavior that treated detached nodes as
visible. This commit makes the default/full `displayCheck` option
now behave correctly in that it treates detached nodes as hidden.

But since many people may depend on the buggy legacy behavior, a
new 'legacy-full' option is added which preserves the legacy
behavior, for now.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ These options apply to all APIs.
 
 ### displayCheck option
 
-Type: `full` | `non-zero-area` | `none` . Default: `full`.
+Type: `full` | `full-with-hidden` | `non-zero-area` | `none` . Default: `full`.
 
 Configures how to check if an element is displayed.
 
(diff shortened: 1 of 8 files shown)


stefcameron:
@craigkovatch @idoros I'm fixing #666 which we had logged a while back after this happened: #664

I'm going to publish a major because I'm dropping all IE support now that IE has been sunset for good, so this is the opportunity to fix the default "full" display check behavior.

I've added a `legacy-full` option that still does the old behavior which folks can use if they need an escape valve for this new major. But going forward, the intention is to eventually drop the legacy behavior completely.

Sounds/looks good?

craigkovatch:
Sounds good thank you

-Craig (mobile)
________________________________

stefcameron:
Thanks for checking it, both of you! Merging.

README.md
```diff
@@ -135,7 +135,7 @@ These options apply to all APIs.
 
 ### displayCheck option
 
-Type: `full` | `non-zero-area` | `none` . Default: `full`.
+Type: `full` | `full-with-hidden` | `non-zero-area` | `none` . Default: `full`.
```
craigkovatch, line 138: I thought the new name was `legacy-full`?

stefcameron: Yes, it's `legacy-full`. Thanks for catching that. Will fix it right away.

PR #837, opened 2022-10-24
adding role=button to selectors

darylshy:
...ADD PR DETAILS HERE...

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ const candidateSelectors = [
   '[contenteditable]:not([contenteditable="false"])',
   'details>summary:first-of-type',
   'details',
+  '[role=button]'
 ];
 const candidateSelector = /* #__PURE__ */ candidateSelectors.join(',');
 


stefcameron:
@darylshy Thanks for wanting to improve Tabbable. Please follow the "PR Checklist" in the description (came from the PR template). Also a bit of explanation, a test, code formatting (run `yarn format`), and a changeset (run `yarn changeset`) would be appreciated...

Does simply giving a role of "button" to an element magically make it tabbable/focusable in all major browsers?

Doesn't look like it to me: https://codesandbox.io/s/hopeful-lalande-0rf9x3

I don't think the `role` attribute has anything to do with making a node tabbable/focusable. Only the `tabindex` attribute can do that for something that is not natively tabbable/focusable like a form element or `<button>`.

stefcameron:
Closing this PR as I don't think this would be a good addition. Please LMK if you think otherwise.

PR #912, opened 2023-01-25
[ftr#905] Fix crash when ownerDocument is null

stefcameron:
Fixes https://github.com/focus-trap/focus-trap-react/issues/905

There's a corner case where a node's `ownerDocument` could be null if
that node is itself the `document`, which could happen under special
circumstances.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>


stefcameron:
Fix confirmed, OK to merge.

PR #924, opened 2023-02-09
[#292] Add support for inert attribute

### LINKED ISSUE #292: support inert attribute
Elements that are marked with [inert](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) or are nested in another `inert` element shouldn't be tabbable/focusable. Since this is just supported behind browsers flags at the moment, maybe tabbable can add support with a default false flag for now.

stefcameron:
Fixes #292

Inert elements, as well as elements inside inert ancestors, will no longer be treated as tabbable/focusable.

Considering this a minor update since inert elements in browsers supporting the inert attribute (which has wide enough support now) would already be non-interactive. So this is tabbable catching to reality rather than tabbable altering behavior.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,15 +1,15 @@
 const candidateSelectors = [
-  'input',
-  'select',
-  'textarea',
-  'a[href]',
-  'button',
-  '[tabindex]:not(slot)',
-  'audio[controls]',
-  'video[controls]',
-  '[contenteditable]:not([contenteditable="false"])',
-  'details>summary:first-of-type',
-  'details',
+  'input:not([inert] *)',
+  'select:not([inert] *)',
+  'textarea:not([inert] *)',
+  'a[href]:not([inert] *)',
+  'button:not([inert] *)',
+  '[tabindex]:not(slot):not([inert] *)',
+  'audio[controls]:not([inert] *)',
+  'video[controls]:not([inert] *)',
+  '[contenteditable]:not([contenteditable="false"]):not([inert] *)',
+  'details>summary:first-of-type:not([inert] *)',
+  'details:not([inert] *)',
 ];
 const candidateSelector = /* #__PURE__ */ candidateSelectors.join(',');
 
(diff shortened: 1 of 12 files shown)


stefcameron:
Unfortunately, FF still does not support `inert`, therefore, henceforth, all FF PR builds will fail as it stands because we don't have a suite of FF-specific tests that don't include use of `inert`...

So I have disabled the requirement for the FF tests to pass.

idoros:
Not sure if this works, but this post suggests Cypress supports conditional test configuration: https://filiphric.com/skip-test-conditionally-with-cypress#test-configuration

src/index.js
```diff
@@ -26,13 +26,29 @@ const getRootNode =
...
+/**
+ * Determines if a node is inert or in an inert ancestor.
+ * @param {Element} node
+ * @param {boolean} [lookUp] If true and `node` is not inert, looks up at ancestors to
+ *  see if any of them are inert. If false, only `node` itself is considered.
+ * @returns {boolean} True if inert itself or by way of being in an inert ancestor.
+ *  False if `node` is falsy.
+ */
+const isInert = function (node, lookUp = true) {
+  return !!(node?.inert || (lookUp && node && isInert(node.parentNode))); // recursive
```
idoros, line 38: Why do you need to lookup with JavaScript instead of matching with CSS match: 

With parent check:
is not inert: `node.matches(':not([inert]), :not([inert] *)')`
is inert:        `node.matches('[inert], :is([inert] *)')`

src/index.js
```diff
@@ -1,15 +1,15 @@
...
-  'textarea',
-  'a[href]',
-  'button',
-  '[tabindex]:not(slot)',
-  'audio[controls]',
-  'video[controls]',
-  '[contenteditable]:not([contenteditable="false"])',
-  'details>summary:first-of-type',
-  'details',
+  'input:not([inert] *)',
```
idoros, line 2: you can add `:not([inert])` to all of these selectors to save the extra JavaScript check on the element itself

stefcameron:
@idoros 
> Not sure if this works, but this post suggests Cypress supports conditional test configuration: https://filiphric.com/skip-test-conditionally-with-cypress#test-configuration

Yes! I had no idea. It was buried in https://github.com/cypress-io/cypress/pull/5346 and https://github.com/cypress-io/cypress-documentation/pull/3209/files but here's the official doc entry: https://docs.cypress.io/guides/references/configuration#Single-test-configuration

Thank you!

src/index.js
stefcameron: Good point, but unfortunately we can't call matches on what I think are the shadow roots. Lots of shadow-related tests fail because of `matches is not a function` errors when I use your suggestion. What I have right now works in all cases.

src/index.js
stefcameron: Thanks for pointing this out. We can eliminate more candidates right off the bat with the addition of `:not([inert])`, which I have done. I didn't use the multi-selector syntax since the exhaustive syntax has broader support (just in case...).

PR #1023, opened 2023-06-08
Make tabbable work with an array of elements

stefcameron:
Per discussions in https://github.com/focus-trap/focus-trap/issues/375 and work in https://github.com/focus-trap/focus-trap/pull/974, we need to:

- [x] Make `tabbable()` work with multiple containers
- [ ] Make `focusable()` work with multiple containers (for API symmetry)
- [ ] Expose `getTabIndex()` API so that focus-trap can use the same fallbacks as tabbable

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -579,19 +579,33 @@ const sortByOrder = function (candidates) {
 const tabbable = function (el, options) {
   options = options || {};
 
+  const containers = Array.isArray(el) ? el : [el];
+
   let candidates;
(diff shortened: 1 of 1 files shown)


stefcameron:
I'll handle the rest of the TODOs in additional commits. For some reason, I'm unable to push to this branch in your fork, though I've pushed commits to forks of contributors before. Not sure what's up there.

PR #1024, opened 2023-06-08
Sort `containers` before querying candidates within them

DaviDevMod:
If the elements passed to `tabbable` are not ordered (by document order) the resulting order of tabbable elements would be incorrect because [sortByOrder](https://github.com/focus-trap/tabbable/blob/a5f9369d1ef93f394a7f8d64e93ceebb4d8a70f0/src/index.js#L545) only sorts the candidates by tab index.

The `.documentOrder` here: 
https://github.com/focus-trap/tabbable/blob/a5f9369d1ef93f394a7f8d64e93ceebb4d8a70f0/src/index.js#L241

Is merely the index of an element within the array of candidates: https://github.com/focus-trap/tabbable/blob/a5f9369d1ef93f394a7f8d64e93ceebb4d8a70f0/src/index.js#L559

Which is also a true indication of the relative document orders when the candidates are queried from a single element or from multiple elements sorted by document order.

I presume the reason for that index is that `Array.prototype.sort()` is [stable only from ECMAScript 2019](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability), it could be renamed to avoid confusion, but I'll leave that up to you.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -603,10 +603,12 @@ const sortByOrder = function (candidates) {
     .concat(regularTabbables);
 };
 
+const byDocumentOrder = (a, b) => (a.compareDocumentPosition(b) & 4 ? -1 : 1);
+
 const tabbable = function (el, options) {
   options = options || {};
 
-  const containers = Array.isArray(el) ? el : [el];
+  const containers = (Array.isArray(el) ? [...el] : [el]).sort(byDocumentOrder);
 
   let candidates;
   if (options.getShadowRoot) {
@@ -642,7 +644,7 @@ const tabbable = function (el, options) {
 const focusable = function (el, options) {
   options = options || {};
 
-  const containers = Array.isArray(el) ? el : [el];
+  const containers = (Array.isArray(el) ? [...el] : [el]).sort(byDocumentOrder);
 
   let candidates;
   if (options.getShadowRoot) {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -603,10 +603,12 @@ const sortByOrder = function (candidates) {
     .concat(regularTabbables);
 };
 
+const byDocumentOrder = (a, b) => (a.compareDocumentPosition(b) & 4 ? -1 : 1);
```
stefcameron, line 606: Could we use the constant property instead of `4` when testing the bitmask? I think that would clarify the intent of the test.

```suggestion
const byDocumentOrder = (a, b) => (a.compareDocumentPosition(b) & Node.DOCUMENT_POSITION_FOLLOWING ? -1 : 1);
```

stefcameron (changes requested):
Thank you for this, very helpful!

src/index.js
DaviDevMod: Of course!

DaviDevMod (commented):
Replace 4 with a more descriptive `Node.DOCUMENT_POSITION_FOLLOWING` constant

stefcameron (approved):
Thank you!

PR #1032, opened 2023-06-14
Many containers

DaviDevMod:
As discussed in https://github.com/focus-trap/focus-trap/issues/375#issuecomment-1591807907, there's no need to sort `containers` and removing nested ones can be done in conjunction with deduping to allow for a simpler logic.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -575,7 +575,7 @@ const isValidShadowRootTabbable = function (shadowHostNode) {
 
 /**
  * @param {Array.<Element|CandidateScope>} candidates
- * @returns {Element[]}
+ * @returns Element[]
  */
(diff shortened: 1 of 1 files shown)


stefcameron:
Thank you for your contributions to my experimental branch. I've made the decision to no longer pursue multi-container support in Tabbable as positive `tabindex` values will be limited to a single-container traps in Focus-trap: https://github.com/focus-trap/focus-trap/issues/375#issuecomment-1602841175

DaviDevMod:
@stefcameron your welcome!

I agree with your decision, the feature it's not worth the hassle. Shadow DOM makes things way too hard here.

Issue #1800, opened 2025-09-18
Idea: using Element checkVisibility() as a new `displayCheck` option


fpapado:
Hi! First of all, thanks for maintaining this library, I know it's a lot of work to keep up with browsers 😌 

I was looking into this library because it is part of our app's dependency tree. Elsewhere in our our app, we have a custom implementation of checking for sequentially and programatically focusable elements, that has run into similar cases as this library.

Our implementation was written with contemporary browsers in mind, so many of the visibility checks are done via [Element checkVisibility()](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility), modulo some shortcuts/fast paths. We've found that it catches most of the edge cases (including `visibility`, `opacity`, as well as the auto state of `content-visibility`).

Would you consider a `displayCheck` option that utilises this method, instead of (or, progressively enhancing) the current bespoke visibility checks? I'd be happy to contribute and help the tests pass 🔧 

[Browser support is quite recent (~March 2024, for Safari 17.4)](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility#browser_compatibility), so my first hunch is to use this as a progressive enhancement.

Let me know what you think!

stefcameron:
@fpapado Thank you! 😄 Yes, I'd be open to a PR that adds using this modern API as an additional `displayCheck` option, provided it's complete with passing test, README (docs) update, and doesn't disturb any of the existing `displayCheck` options.

fpapado:
Pefect! I'll work on it this and next week ✍️

PR #1818, opened 2025-10-01
feat: add 'full-native' `displayCheck` option, utilising Element#checkVisibility

fpapado:
...ADD PR DETAILS HERE...

<details>
<br/>

- Web APIs introduced have __deep__ browser coverage, including Safari (often very late to adopt new APIs).
- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -404,6 +404,9 @@ const isHidden = function (node, { displayCheck, getShadowRoot }) {
   // NOTE: we check this regardless of `displayCheck="none"` because this is a
   //  _visibility_ check, not a _display_ check
   if (getComputedStyle(node).visibility === 'hidden') {
+    // TODO(fpapado): check whether this makes sense in a 'checkVisibility'
+    // world; we might be able to avoid the style calculation (at least, on our
+    // end) here
     return true;
   }
 
@@ -411,6 +414,19 @@ const isHidden = function (node, { displayCheck, getShadowRoot }) {
   const nodeUnderDetails = isDirectSummary ? node.parentElement : node;
   if (matches.call(nodeUnderDetails, 'details:not([open]) *')) {
     return true;
+  }
+
+  if (!displayCheck || displayCheck === 'full') {
+    if ('checkVisibility' in node) {
+      const visible = node.checkVisibility({
+        contentVisibilityAuto: true,
+        opacityProperty: true,
+        visibilityProperty: true,
+      });
+      return !visible;
+    }
+    // if checkVisibility is not supported, we fall through to the other
+    // checks
   }
 
   if (
--- a/src/index.js
+++ b/src/index.js
@@ -398,14 +398,20 @@ const isZeroArea = function (node) {
   return width === 0 && height === 0;
 };
 const isHidden = function (node, { displayCheck, getShadowRoot }) {
-  if (!displayCheck || displayCheck === 'full') {
-    // Specifically for the full check (or the default), we take a fast path via
-    // Element#checkVisibility
+  if (displayCheck === 'full-native') {
     if ('checkVisibility' in node) {
+      // Chrome >= 105, Edge >= 105, Firefox >= 106, Safari >= 17.4
+      // @see https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility#browser_compatibility
       const visible = node.checkVisibility({
         contentVisibilityAuto: true,
         opacityProperty: true,
         visibilityProperty: true,
+        // These two are aliases for opacityProperty and visibilityProperty.
+        // Contemporary browsers support both. However, these aliases have wider
+        // browser support (Chrome >= 105 and Firefox >= 106, vs. Chrome >= 121
+        // and Firefox >= 122), so we include them anyway.
+        checkOpacity: true,
+        checkVisibilityCSS: true,
       });
       return !visible;
     }
--- a/src/index.js
+++ b/src/index.js
@@ -398,6 +398,20 @@ const isZeroArea = function (node) {
   return width === 0 && height === 0;
 };
 const isHidden = function (node, { displayCheck, getShadowRoot }) {
+  if (!displayCheck || displayCheck === 'full') {
+    // Specifically for the full check (or the default), we take a fast path via
+    // Element#checkVisibility
+    if ('checkVisibility' in node) {
+      const visible = node.checkVisibility({
+        contentVisibilityAuto: true,
+        opacityProperty: true,
+        visibilityProperty: true,
+      });
+      return !visible;
+    }
+    // Fall through to manual visibility checks
+  }
+
   // NOTE: visibility will be `undefined` if node is detached from the document
   //  (see notes about this further down), which means we will consider it visible
   //  (this is legacy behavior from a very long way back)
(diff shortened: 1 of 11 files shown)


test/e2e/e2e.helpers.js
```diff
@@ -1,8 +1,8 @@
 import { appendHTMLWithShadowRoots } from '../shadow-root-utils';
 
-export function setupTestWindow(done) {
+export function setupTestDocument(done) {
   cy.visit('./cypress/test-sandbox.html');
-  cy.window().then(done);
+  cy.document().then(done);
```
fpapado, line 5: I am not sure why, but even on the main branch, local E2E tests simply would not pass/run correctly with the existing `setupTestWindow`. [Since Cypress' docs seem to recommend `cy.document()` as a way to get a handle on the current window document](https://docs.cypress.io/api/commands/document#Yields), I used that instead of going through the window. 

(I'm significantly more familiar with Playwright, so I do not know if this is the most idiomatic way to do this)

test/e2e/focusable.cy.js
```diff
@@ -25,160 +25,169 @@ describe('focusable', () => {
   });
 
   describe('example fixtures', () => {
-    [undefined, 'full', 'legacy-full'].forEach((displayCheck) => {
```
fpapado, line 28: all that change in tests cases is the addition of 'full-native' as a parameter. The formatter exploded the diff a bit 😁

src/index.js
```diff
@@ -398,6 +398,26 @@ const isZeroArea = function (node) {
...
+        // Contemporary browsers support both. However, these aliases have wider
+        // browser support (Chrome >= 105 and Firefox >= 106, vs. Chrome >= 121
+        // and Firefox >= 122), so we include them anyway.
+        checkOpacity: true,
+        checkVisibilityCSS: true,
+      });
+      return !visible;
+    }
+    // Fall through to manual visibility checks
+  }
```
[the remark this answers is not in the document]
fpapado: Oh this is a great point actually! The fallback still checks for the default or the other 'full' variants. This would be caught if we were testing both in supporting and unsupporting browsers. Maybe worth adding a matrix for it (I can do it in this PR).

fpapado:
@stefcameron, I opened this PR after discussing in https://github.com/focus-trap/tabbable/issues/1800. The GitHub UI does not allow me to assign reviewers on this repository, so sending this ping just in case ✉️

stefcameron:
@fpapado Thank for the PR, saw it! I will review within the next couple of days.

stefcameron:
@fpapado Had a look! Looks pretty solid. But the one thing I'm not seeing, or missing seeing, is at least one test that verifies the new `full-native` option is behaving as expected in contrast to `full`, and also the fallback to `full` when `checkVisibility()` isn't defined.

I'd like to see one test that shows `full-native` finds something that `full` doesn't, or the opposite, which ever, and another that shows `full-native` drops to `full` when the API isn't available.

Is that possible?

fpapado:
> @fpapado Had a look! Looks pretty solid. But the one thing I'm not seeing, or missing seeing, is at least one test that verifies the new `full-native` option is behaving as expected in contrast to `full`, and also the fallback to `full` when `checkVisibility()` isn't defined.
> 
> I'd like to see one test that shows `full-native` finds something that `full` doesn't, or the opposite, which ever, and another that shows `full-native` drops to `full` when the API isn't available.
> 
> Is that possible?

Yeah, these are both doable. I added a test for something that `full-native` supports (`content-visibility: hidden` parent) now in 5bf6ca729401e0f8c9c1a595f25e7fd20087ae68.

Testing the fallback in end-to-end proved slightly cumbersome, because both Chrome and Firefox in the existing CI docker image support `checkVisibility`. [One of Cypress' Node 16 images _do_ have matching browser versions](https://hub.docker.com/r/cypress/browsers/tags?name=105), but I wanted to avoid going further into unsupported Node versions on CI. My personal hunch would be to create docker images with the specific versions of browsers that we'd want to test, but that's a bit of a rabbit hole.

Anyway, to work around that, I tested the fallback in unit tests, since JSDOM in this case conveniently does not support `checkVisibility`. The relevant change is at dd9f4fe7ae1bc33a8f2222610ec722b6e7864c16, with a check to ensure that the assumptions about the environment are correct.

stefcameron:
@fpapado Just one last small request, if that works with you. Also FF failed in CI: https://github.com/focus-trap/tabbable/actions/runs/18322680365/job/52302880471?pr=1818

fpapado:
> @fpapado Just one last small request, if that works with you. Also FF failed in CI: https://github.com/focus-trap/tabbable/actions/runs/18322680365/job/52302880471?pr=1818

The FF failure sent me down a small rabbit hole, when it comes to what Firefox considers visible and focusable 🔍  

After some bisecting ([full instructions in this repository](https://github.com/fpapado/firefox-checkVisibility-with-contentVisibilityHidden), I used Playwright since I find it much easier to get a specific browser version on demand, compared to juggling docker image builds), it turns out that:

- Firefox 106, which the CI uses, supports `checkVisibility` and the `checkVisibilityCSS` option, but considers elements with a `content-visibility: hidden` parent to be visible, and also focusable/tabbable
- Firefox 125 is the first version that considers such elements both invisible and not focusable/tabbable.

Since the logic is internally consistent, I don't think it is a problem, because `tabbable` follows exactly what the browser (seems to be) doing. So, **only in the tests**, I added a check for Firefox < 125 and made the assertions accurate in 353c86acc39b9806e21288504af4f460023aae7d. Let me know if this makes sense!

stefcameron:
> > @fpapado Just one last small request, if that works with you. Also FF failed in CI: https://github.com/focus-trap/tabbable/actions/runs/18322680365/job/52302880471?pr=1818
> 
> The FF failure sent me down a small rabbit hole, when it comes to what Firefox considers visible and focusable 🔍
> 
> After some bisecting ([full instructions in this repository](https://github.com/fpapado/firefox-checkVisibility-with-contentVisibilityHidden), I used Playwright since I find it much easier to get a specific browser version on demand, compared to juggling docker image builds), it turns out that:

Thank you for investigating, much appreciated!

> Since the logic is internally consistent, I don't think it is a problem, because `tabbable` follows exactly what the browser (seems to be) doing. So, **only in the tests**, I added a check for Firefox < 125 and made the assertions accurate in [353c86a](https://github.com/focus-trap/tabbable/commit/353c86acc39b9806e21288504af4f460023aae7d). Let me know if this makes sense!

Yes, I think that's OK.

stefcameron:
Thanks for trying to fix the test earlier. FF is still giving some grief... 😬

fpapado:
> Thanks for trying to fix the test earlier. FF is still giving some grief... 😬

Hello! I took a small break due to work, but I think I figured it out now in 22d0f74b490a92e26d852503e04c6ab991488b00. Turns out that `Cypress.browser.name` is lowercase, so the FF < 125 detection was not working as intended 😅 Let's see what CI thinks!

stefcameron:
> > Thanks for trying to fix the test earlier. FF is still giving some grief... 😬
> 
> Hello! I took a small break due to work, but I think I figured it out now in [22d0f74](https://github.com/focus-trap/tabbable/commit/22d0f74b490a92e26d852503e04c6ab991488b00). Turns out that `Cypress.browser.name` is lowercase, so the FF < 125 detection was not working as intended 😅 Let's see what CI thinks!

No worries, thank you for getting back to it! We're 🟢 in CI so I'm going to merge this and get it published!

stefcameron (approved):

stefcameron:
@all-contributors add @fpapado for code, test

stefcameron:
published with https://github.com/focus-trap/tabbable/pull/1848

PR #995, opened 2023-05-03
[#982] Pin nwsapi downstream jsdom dependency to v2.2.2

stefcameron:
Fixes #982

Tested locally by cloning this repo and then:
```
$ rm package-lock.json
$ npm install # regenerates package-lock.json
$ npm test # fails because of nwsapi issues post-v2.2.2
# edit package.json to add override clause
$ npm install # results in nwsapi@2.2.2
$ npm test # succeeds
```

<details>
<summary>PR Checklist</summary>
<br/>

__Please leave this checklist in your PR.__

- Source changes maintain stated browser compatibility.
- Issue being fixed is referenced.
- Unit test coverage added/updated.
- E2E test coverage added/updated.
- Typings added/updated.
- Changes do not break SSR:
  - Careful to test `typeof document/window !== 'undefined'` before using it in code that gets executed on load.
- README updated (API changes, instructions, etc.).
- Changes to dependencies explained.
- Changeset added (run `npm run changeset` locally to add one, and follow the prompts).
  - EXCEPTION: A Changeset is not required if the change does not affect any of the source files that produce the package bundle. For example, demo changes, tooling changes, test updates, or a new dev-only dependency to run tests more efficiently should not have a Changeset since it will not affect package consumers.

</details>

PR #1879, opened 2025-11-30
feat: use native `:not([inert] *)` selector to filter out elements in inert subtrees

### LINKED ISSUE #997: Remove nwsapi v2.2.2 override once bug is fixed
Relates to #982

Would revert #995 

Might want to make sure package-lock.json is also now using the nwsapi version that has the fix instead of being still configured for v2.2.2 (which is why #982 isn't visible unless you regenerate the lock file and NPM picks the current-latest problematic version of nwsapi).

fpapado:
Closes #997 

Hi @stefcameron 👋 While working on #1818 and tweaking the code, I noticed that JSDOM has started supporting `:not([inert] *)` again*.  

Edit: aha, you also spotted that in https://github.com/focus-trap/tabbable/issues/997#issuecomment-3592077879 😅 

So in this PR I am reintroducing this selector, which should make the tabbable checks a bit more efficient, and the code a bit cleaner (there's one less check to do manually). The unit/component tests (with JSDOM 26.1.0) pass without change.

*It is a bit unclear to me whether the reason is `nwsapi` fixing its parsing, or the fact that JSDOM is using a different CSS selector engine. I can bisect JSDOM versions a bit, to see which one it is.

<details>
<br/>

- Web APIs introduced have __deep__ browser coverage, including Safari (often very late to adopt new APIs).
- Typings added/updated.
- Changes do not break SSR:

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored a fast-path CSS selector check so elements inside inert subtrees are correctly excluded, improving focusability detection and compatibility with newer JSDOM versions.

* **Chores**
  * Simplified inert-handling logic for more robust element detection and removed an unnecessary package override.

* **Tests**
  * Streamlined end-to-end tests to use a helper-based ID comparison for verifying focusable elements.

* **Documentation**
  * Added a minor release note describing inert subtree selector behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -50,10 +50,11 @@ const isInert = function (node, lookUp = true) {
     inert ||
     (lookUp &&
       node &&
-      // closest is only defined on Element, but does not exist on shadow roots,
-      // so we fall back to manual lookups upward in case it is not defined.
-      ((typeof node.closest === 'function' && node.closest('[inert]')) ||
-        isInert(node.parentNode)));
+      // closest does not exist on shadow roots, so we fall back to a manual
+      // lookup upward, in case it is not defined.
+      (typeof node.closest === 'function'
+        ? node.closest('[inert]')
+        : isInert(node.parentNode)));
 
   return result;
 };
--- a/src/index.js
+++ b/src/index.js
@@ -1,20 +1,17 @@
 // NOTE: separate `:not()` selectors has broader browser support than the newer
 //  `:not([inert], [inert] *)` (Feb 2023)
-// CAREFUL: JSDom does not support `:not([inert] *)` as a selector; using it causes
-//  the entire query to fail, resulting in no nodes found, which will break a lot
-//  of things... so we have to rely on JS to identify nodes inside an inert container
 const candidateSelectors = [
-  'input:not([inert])',
-  'select:not([inert])',
-  'textarea:not([inert])',
-  'a[href]:not([inert])',
-  'button:not([inert])',
-  '[tabindex]:not(slot):not([inert])',
-  'audio[controls]:not([inert])',
-  'video[controls]:not([inert])',
-  '[contenteditable]:not([contenteditable="false"]):not([inert])',
-  'details>summary:first-of-type:not([inert])',
-  'details:not([inert])',
+  'input:not([inert]):not([inert] *)',
+  'select:not([inert]):not([inert] *)',
+  'textarea:not([inert]):not([inert] *)',
+  'a[href]:not([inert]):not([inert] *)',
+  'button:not([inert]):not([inert] *)',
+  '[tabindex]:not(slot):not([inert]):not([inert] *)',
+  'audio[controls]:not([inert]):not([inert] *)',
+  'video[controls]:not([inert]):not([inert] *)',
+  '[contenteditable]:not([contenteditable="false"]):not([inert]):not([inert] *)',
+  'details>summary:first-of-type:not([inert]):not([inert] *)',
+  'details:not([inert]):not([inert] *)',
 ];
 const candidateSelector = /* #__PURE__ */ candidateSelectors.join(',');
 
@@ -52,6 +49,7 @@ const isInert = function (node, lookUp = true) {
   // CAREFUL: JSDom does not appear to support certain selectors like `:not([inert] *)`
   //  so it likely would not support `:is([inert] *)` either...
   const result = inert || (lookUp && node && isInert(node.parentNode)); // recursive
+  // const result = inert || (lookUp && node && isInert(node.closest('[inert]'))); // recursive
 
   return result;
 };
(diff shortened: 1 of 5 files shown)


test/e2e/isFocusable.cy.js
```diff
@@ -428,9 +429,9 @@ describe('isFocusable', () => {
         container.innerHTML = fixtures.inert;
         document.body.append(container);
 
-        for (const child of container.children) {
-          expect(isFocusable(child)).to.eql(false);
-        }
+        const focusables = Array.from(container.children).filter(isFocusable);
+
+        expect(getIdsFromElementsArray(focusables)).to.eql([]);
```
fpapado: Not related to the change, but it makes debugging easier (you get a complete list of `id`s in the diff, instead of a single true/false diff). Other tests also show lists ids in the diff, so it seemed more consistent.

src/index.js
```diff
@@ -1,20 +1,17 @@
...
-  'textarea:not([inert])',
-  'a[href]:not([inert])',
-  'button:not([inert])',
-  '[tabindex]:not(slot):not([inert])',
-  'audio[controls]:not([inert])',
-  'video[controls]:not([inert])',
-  '[contenteditable]:not([contenteditable="false"]):not([inert])',
-  'details>summary:first-of-type:not([inert])',
-  'details:not([inert])',
+  'input:not([inert]):not([inert] *)',
```
fpapado: I still used `:not([inert]):not([inert] *)` instead of `:not([inert], [inert] *)`. The former seems less likely to go wrong in custom CSS selector engines, and anyway I'd expect them to be equivalent in terms of performance.

stefcameron: I'm OK with the double `:not` instead of the `:not` list, but I'm not sure we can safely replace the 3 lines you've deleted with the single replacement. `contenteditable` is not an attribute on an `input`; you can put that on any element, e.g. to make a `div` editable. And `details` is not an input field either.

stefcameron:
@fpapado Thanks for the follow-up here. I think to truly see that the issue is fixed, we need to revert #995 and see that it works.

fpapado:
> @fpapado Thanks for the follow-up here. I think to truly see that the issue is fixed, we need to revert #995 and see that it works.

Ah yes, I missed that! I removed the override now. `npm ls nwsapi` says that version `2.2.22` (instead of `2.2.2`) is used:

```shell
tabbable@6.3.0 /Users/fotis/oss/tabbable
└─┬ jest-environment-jsdom@30.2.0
  └─┬ jsdom@26.1.0
    └── nwsapi@2.2.22
```

Versions down to `2.2.16` (the lowest supported by JSDOM 26) seem to pass tests as well.

stefcameron:
> > @fpapado Thanks for the follow-up here. I think to truly see that the issue is fixed, we need to revert #995 and see that it works.
> 
> Ah yes, I missed that! I removed the override now. `npm ls nwsapi` says that version `2.2.22` (instead of `2.2.2`) is used:
> 
> ```shell
> tabbable@6.3.0 /Users/fotis/oss/tabbable
> └─┬ jest-environment-jsdom@30.2.0
>   └─┬ jsdom@26.1.0
>     └── nwsapi@2.2.22
> ```
> 
> Versions down to `2.2.16` (the lowest supported by JSDOM 26) seem to pass tests as well.

Thanks for this!

You may have missed this comment? https://github.com/focus-trap/tabbable/pull/1879#discussion_r2575330139

src/index.js
```diff
@@ -49,9 +46,7 @@ const isInert = function (node, lookUp = true) {
   // NOTE: this could also be handled with `node.matches('[inert], :is([inert] *)')`
   //  if it weren't for `matches()` not being a function on shadow roots; the following
   //  code works for any kind of node
-  // CAREFUL: JSDom does not appear to support certain selectors like `:not([inert] *)`
-  //  so it likely would not support `:is([inert] *)` either...
-  const result = inert || (lookUp && node && isInert(node.parentNode)); // recursive
+  const result = inert || (lookUp && node && node.closest('[inert]'));
```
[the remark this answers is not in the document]
fpapado: I was originally confused by Coderabbit's comment, because the `node` argument was typed as `Element` in the JSDoc (instead of `Node`, as it mentioned). But it makes sense that it is Node, given also the existing code comments about `matches` not being defined on shadow roots 👍 

I fixed this now in: a01931deb8b1221ff02a4581d7395f731e8bd796

I kept the manual lookup upward in case `closest()` is not defined, I don't think it hurts, and if the Node's parent is an Element, then the manual lookup will hit the fast path via `closest()` again.

stefcameron (approved):
I think we're good to go on this one! Thank you for help with this, @fpapado

stefcameron:
@all-contributors add @fpapado for code

stefcameron:
I just noticed I had not published this yet. I somehow missed the PR to trigger publishing. Published in 6.4.0 with https://github.com/focus-trap/tabbable/pull/1899

stefcameron (approved):

